PR #23, opened 2024-01-24
Feature 18 on call lambda

petri-nykanen:
This PR enables loading on-call data and check for on-call person of the week.


serverless.ts
```diff
@@ -18,6 +19,8 @@ import listTimeEntriesHandler from "src/functions/forecast/list-time-entries";
 import listProjectSprintsHandler from "src/functions/forecast/list-project-sprints";
 import sendDailyMessage from "@functions/meta-assistant/send-daily-message";
 import sendWeeklyMessage from "@functions/meta-assistant/send-weekly-message";
+import updatePaidHandler from "src/functions/update-paid";
+
```
```suggestion
```

serverless.ts
```diff
@@ -1,8 +1,9 @@
+import type { AWS } from "@serverless/typescript";
+import loadOnCallDataHandler from "@functions/load-on-call-data";
+import weeklyCheckHandler from "@functions/weekly-check/";
```
Jdallos: move with other function imports

src/functions/load-on-call-data/handler.ts
```diff
@@ -0,0 +1,43 @@
...
+    
+    const year = parseInt(queryStringParameters.year);
+    if (!year || year < 2020 || year > new Date().getFullYear()) {
+        throw new Error("Invalid year");
+    }
+
+    const s3 = new S3();
+
+    const bucket = Config.get().onCall.bucketName;
+    const nameMap = await S3Utils.loadJson<{ [key: string]: string } >(s3, bucket, "name-map.json") || {};
```
Jdallos: Does this always take name-map.json, if so could this be a const with appropriate name, it's not clear to me what this file refers to at the moment.

src/functions/load-on-call-data/handler.ts
```diff
@@ -0,0 +1,43 @@
...
+    const data = await S3Utils.loadJson<OnCallEntry[]>(s3, bucket, `${year}.json`);
+    const paidData = await S3Utils.loadJson<PaidData>(s3, bucket, "paid.json") || {};
```
Jdallos: Same here as with name-map.json

src/functions/load-on-call-data/handler.ts
```diff
@@ -0,0 +1,43 @@
...
+    const year = parseInt(queryStringParameters.year);
+    if (!year || year < 2020 || year > new Date().getFullYear()) {
+        throw new Error("Invalid year");
+    }
+
+    const s3 = new S3();
+
+    const bucket = Config.get().onCall.bucketName;
+    const nameMap = await S3Utils.loadJson<{ [key: string]: string } >(s3, bucket, "name-map.json") || {};
+    const data = await S3Utils.loadJson<OnCallEntry[]>(s3, bucket, `${year}.json`);
```
Jdallos: To keep this consistent with the other file names I have suggested being a const, perhaps this could be declared as a variable populated with `${year}.json`. Not an essential change but might make it a little more readable with the other changes.

src/functions/load-on-call-data/handler.ts
```diff
@@ -0,0 +1,43 @@
...
+    const paidData = await S3Utils.loadJson<PaidData>(s3, bucket, "paid.json") || {};
+
+    if (!data) {
```
Jdallos: Should this also handle if !nameMap or !paidData or is this not necessary?

src/functions/load-on-call-data/handler.ts
Jdallos: Also if you disagree with these naming suggestions feel free to ignore :D

src/functions/update-paid/handler.ts
```diff
@@ -0,0 +1,42 @@
+import { S3 } from "aws-sdk"
+import { PaidData } from "../../types"
+import S3Utils from "@libs/s3-utils";
+import Config from "src/app/config";
+import { middyfy } from "@libs/lambda";
+
+/**
+ * Request body interface
+ */
+interface RequestBody {
```
Jdallos: Could this be renamed to be specific for the function (in the future it might make sense for us to export these interfaces from a types file, but this is not necessary for now),

src/functions/update-paid/handler.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/**
+ * Lambda method for updating paid data
+ * 
+ * @param event event
+ */
+export const updatePaidHandler = async (event: { body: string }) => {
+    const requestBody: RequestBody = JSON.parse(event.body);
+    const { year, week, paid } = requestBody;
```
Jdallos: Do we need some check for paid as you have for year and week?

src/functions/update-paid/handler.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    if (!week || week < 1 || week > 53) {
+        throw new Error("Invalid week");
+    }
+    
+    const s3 = new S3();
+    const bucket = Config.get().onCall.bucketName;
+    
+    const paidData = await S3Utils.loadJson<PaidData>(s3, bucket, "paid.json") || {};
```
Jdallos: Use a const for the file name?

src/functions/update-paid/handler.ts
```diff
@@ -0,0 +1,42 @@
...
+    paidData[year] = paidData[year] || {};
```
Jdallos: What is this line doing? Is it actually checking if paidData[year] is defined otherwise setting it to a {}, in which case could we be more explicit e.g. if (!paidData[year]) paidData[year] = {}

src/functions/update-paid/handler.ts
```diff
@@ -0,0 +1,42 @@
...
+    paidData[year][week] = paid;
+    
+    await S3Utils.saveJson(s3, bucket, "paid.json", paidData);
```
Jdallos: File name from const?

src/functions/update-paid/handler.ts
```diff
@@ -0,0 +1,42 @@
...
+};
```
Jdallos:
```suggestion
};

```

src/functions/update-paid/index.ts
```diff
@@ -0,0 +1,6 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [  ],
```
Jdallos: Is this function not ready, if there is no event can this function be invoked, in which case do we need to merge this with develop at this point?

src/functions/weekly-check/handler.ts
```diff
@@ -0,0 +1,126 @@
...
+import { OnCallEntry } from "../../types"
+import S3Utils from "@libs/s3-utils";
+import { middyfy } from "@libs/lambda";
+import Config from "src/app/config";
+
+/**
+ * Schedule from Splunk
+ */
+interface Schedule {
+    team: {
```
Jdallos: Can you make sure indentation is set to 2 spaces for all files please.

src/functions/weekly-check/handler.ts
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Resolve next week from schedule
+ * 
+ * @param schedule schedule
+ * @param nextThursday next Thursday
+ * @param policyName name of policy
+ * @returns week from schedule
+ */
+const getNextWeekFromSchedule = (schedule: Schedule, nextThursday: DateTime, policyName: string) => {
+    const scheduleSchedule = schedule.schedules.find(shedule => shedule.policy.name === policyName)?.schedule[0];
```
Jdallos: Can this name be more descriptive

src/functions/weekly-check/handler.ts
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Lambda for checking who is on call this week
+ * 
+ * @param event event
+ */
+export const weeklyCheckHandler = async () => {
+    const { SPLUNK_SCHEDULE_POLICY_NAME, SPLUNK_TEAM_ONCALL_URL, SPLUNK_API_ID, SPLUNK_API_KEY } = process.env;
+    if (!SPLUNK_SCHEDULE_POLICY_NAME || !SPLUNK_TEAM_ONCALL_URL || !SPLUNK_API_ID || !SPLUNK_API_KEY) {
+        throw new Error('Missing environment variables');
+    }
```
Jdallos: Can these come through the app/config with the rest of the envs?

src/functions/weekly-check/handler.ts
```diff
@@ -0,0 +1,126 @@
...
+    
+    const schedule = await (await fetch(`${splunkTeamOnCallUrl}/schedule?daysForward=4&daysSkip=3`, {
+        headers: {
+            'X-VO-Api-Id': SPLUNK_API_ID,
+            'X-VO-Api-Key': SPLUNK_API_KEY,
+            'Accept': 'application/json'
+        } 
+    })).json() as Schedule;
+
+    const nextThursday = DateTime.now().plus({ days: 4 });
```
Jdallos: This seems a bit odd, could the name be improved? I assume this lambda only runs on a Sunday?

src/functions/weekly-check/handler.ts
```diff
@@ -0,0 +1,126 @@
...
+    const nextWeek = getNextWeekFromSchedule(schedule, nextThursday, SPLUNK_SCHEDULE_POLICY_NAME);
+    if (!nextWeek) {
+        throw new Error("No next week");
```
Jdallos: This error seems to be worded a bit string, how about "Next week not found"?

src/functions/weekly-check/handler.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    const fileName = `${nextThursday.year}.json`;
+    const s3 = new S3();
+    const bucket = Config.get().onCall.bucketName;
+    const yearJson = (await S3Utils.loadJson<OnCallEntry[]>(s3, bucket, fileName)) || [];
+    
+    const index = yearJson.findIndex(entry => entry.Week == nextWeek.week);
```
Jdallos: More descriptive name than index? Is it selected week?

src/functions/weekly-check/index.ts
```diff
@@ -0,0 +1,6 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [  ],
```
Jdallos: Same again here as for the update paid function, if there is no event how is this function called?

src/libs/mockData.json
Jdallos, line 1: Unless you need it I think this file can be removed

src/libs/s3-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+            }
+        
+            return JSON.parse(body.toString());
+        } catch (e: any) {
+            if (e.code === "NoSuchKey") {
+                return null;
+            }
+            
+            throw e;
+        }
```
Jdallos: This looks a little suspicious but I'm happy if you are, is this how it was in the previous code?

src/libs/s3-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+     * @param data data to be saved as JSON
+     */
+    export const saveJson = async <T> (s3: S3, bucket: string, key: string, data: T): Promise<void> => {
+        await s3.putObject({
+            Bucket: bucket,
+            Key: key,
+            Body: JSON.stringify(data)
+        }).promise();
+    }
+
```
Jdallos, line 36:
```suggestion
```

src/functions/weekly-check/handler.ts
```diff
@@ -0,0 +1,126 @@
...
+import { DateTime } from "luxon";
+import { OnCallEntry } from "../../types"
+import S3Utils from "@libs/s3-utils";
+import { middyfy } from "@libs/lambda";
+import Config from "src/app/config";
+
+/**
+ * Schedule from Splunk
+ */
+interface Schedule {
```
Jdallos: Could this interface be moved into the types/index.ts?

Jdallos (changes requested):
Changes requested

src/libs/s3-utils.ts
petri-nykanen: This code was indeed taken from the old calendar (on-call-ui), with minor modifications by Antti to how the files are fetched from the bucket, as that was done differently in the SST framework. I'll leave it as it is for now.

src/functions/load-on-call-data/handler.ts
petri-nykanen: It's a good idea. The return value is on-call weeks of that year, which adds person name and paid status props from the other two files. The added conditions could prevent cases where on-call data exists, but not the other two files.

src/functions/weekly-check/handler.ts
petri-nykanen: I couldn't think much better of a name, I thought about "fourDaysFromSunday" but this could sound as cryptic. I opted to keep this variable name and update its description in getNextWeekFromSchedule JSDOC, where it is used as a parameter.

src/functions/on-call/load-on-call-data/mockData.json
Jdallos, line 1: Could you remove all of the mockData.json files you have added within the functions. There is one in meta-assistant/debug/mock.json if you need one for use testing locally later on but otherwise we end up with too many mockData.json files scattered throughout the project.

src/functions/on-call/update-paid/mockData.json
Jdallos, line 1: Remove

src/functions/on-call/weekly-check/mockData.json
Jdallos, line 1: remove

src/types/on-call/index.ts
```diff
@@ -0,0 +1,122 @@
+/**
+ * Date range type
+ */
+export type DateRange = {
+  start_date: string,
+  end_date: string
+};
```
Jdallos: This is not used in the oncall lambdas, can you leave this in the types/index.ts and I'll get the trainees to create the appropriate folder structure.

src/types/on-call/index.ts
```diff
@@ -0,0 +1,122 @@
...
+  access_token: string;
+  email?: string;
+  expires_in?: number;
+  refresh_token?: string;
+  refresh_expires_in?: number;
+  firstName?: string;
+  lastName?: string;
+  userId?: string;
+  roles?: string[];
+}
```
Jdallos: Looks like this isn't used, if you agree please remove

src/types/on-call/index.ts
```diff
@@ -0,0 +1,122 @@
...
+    apiKey: string,
+    apiUrl: string
+  },
+  splunkApi: {
+    apiKey: string
+    apiId: string,
+    teamOnCallUrl: string,
+    schedulePolicyName: string
+  }
+}
```
Jdallos: As this isn't just used by the oncall lambdas, it should also be in the general types/index.ts

Jdallos (changes requested):
Couple of changes and can then be approved.

Jdallos (approved):

PR #27, opened 2024-02-26
bugfix-26-project-and-allocation-handler-bugs

Kseniia-Koshkina:
Changed filter-utils to include projects with endDate null in listProjectsHandler.

### DIFF
--- a/src/functions/forecast/list-project-sprints/handler.ts
+++ b/src/functions/forecast/list-project-sprints/handler.ts
@@ -1,3 +1,4 @@
+// TODO: The related forecast endpoint is no longer used and so if this is still not needed when sprint view UI development is complete, it should be removed.
 import { ValidatedEventAPIGatewayProxyEvent } from "@libs/api-gateway";
 import { middyfy } from "@libs/lambda";
 import { CreateForecastApiService, ForecastApiService } from "src/apis/forecast-api-service";
@@ -40,7 +41,7 @@ const listProjectSprints = async (api: ForecastApiService, parameters: ListProje
     };
   });
 }
-// TODO: The related forecast endpoint is no longer used and so if this is still not needed when sprint view UI development is complete,
+
 /**
  * Lambda for listing Forecast project sprints
  * 
--- a/src/functions/forecast/list-project-sprints/handler.ts
+++ b/src/functions/forecast/list-project-sprints/handler.ts
@@ -40,7 +40,7 @@ const listProjectSprints = async (api: ForecastApiService, parameters: ListProje
     };
   });
 }
-
+// TODO: The related forecast endpoint is no longer used and so if this is still not needed when sprint view UI development is complete,
 /**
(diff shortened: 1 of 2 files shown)


src/functions/forecast/list-project-sprints/handler.ts
```diff
@@ -40,7 +40,7 @@ const listProjectSprints = async (api: ForecastApiService, parameters: ListProje
     };
   });
 }
-
+// TODO: The related forecast endpoint is no longer used and so if this is still not needed when sprint view UI development is complete,
```
Jdallos: add to the end "it should be removed".
I also think this should be moved to the first line of the file.

src/libs/filter-utils.ts
```diff
@@ -10,23 +10,25 @@ export namespace FilterUtilities {
    * @returns If specified dates were null or not, if specified date parameters were between Forecast dates or if currentDate is between Forecast dates
    */
   export const filterByDate = (dateRange: DateRange, currentDate: Date, parameters: { startDate?: Date, endDate?: Date }): boolean => {
-    if (dateRange.start_date === null || dateRange.end_date === null) {
+    if (dateRange.start_date === null) {
       return false;
     }
-
+    const start_date = new Date(dateRange.start_date);
+    let end_date = new Date(dateRange.end_date);
```
Jdallos: I'm not sure why DateRange was set up with keys using snake_case but lets at least use camelCase for these two variables here

Jdallos: I would add a line break after these variables for code clarity

src/libs/filter-utils.ts
```diff
@@ -10,23 +10,25 @@ export namespace FilterUtilities {
...
+    if (dateRange.end_date === null) end_date = currentDate;
     if (parameters.startDate) {
-      if (parameters.startDate <= new Date(dateRange.start_date)) {
+      if (parameters.startDate <= start_date) {
```
Jdallos: Rather than these nested ifs I believe you could use condition:
(parameters.startDate && parameters.startDate <= start_date)

Perhaps you could even use (parameters?.startDate <= start_date)
but you should check the filter still works as intended

src/libs/filter-utils.ts
```diff
@@ -10,23 +10,25 @@ export namespace FilterUtilities {
...
         return false;
       }
-    } else if (currentDate <= new Date(dateRange.start_date)) {
+    } else if (currentDate <= start_date) {
       return false;
     }
 
     if (parameters.endDate) {
-        if (parameters.endDate >= new Date(dateRange.end_date)) {
+        if (parameters.endDate >= end_date) {
```
Jdallos: Same as above, can we avoid the nested if condtions?

Jdallos (changes requested):
Changes requested

Jdallos (approved):

PR #29, opened 2024-02-27
date type fixed

### LINKED ISSUE #28: MetaAssistant Time utils refactoring
Changing date strings from getPreviousTwoWorkdays and other methods to use Date type rather than string.

DZotoff:


src/meta-assistant/timebank/timebank-api.ts
```diff
@@ -40,8 +41,10 @@ namespace TimebankApi {
    * @returns Array of time entries for person
    * @param accessToken string for timebank accessToken
    */
-  export const getDailyEntries = async (id: number, before: string, after: string, accessToken: string): Promise<DailyEntry> => {
+  export const getDailyEntries = async (id: number, beforeDate: DateTime, afterDate: DateTime, accessToken: string): Promise<DailyEntry> => {
     try {
+      let before = beforeDate.toISODate();
+      let after = afterDate.toISODate();
```
Jdallos: Any reason these can't be declared as const, as a general rule variables should be const unless there is a specific reason for something to be a let.

Jdallos (changes requested):
Changes requested, ensure the function is still invoking successfully

Jdallos (approved):

PR #31, opened 2024-03-14
Home lambdas error handling and filtering

Kseniia-Koshkina:


src/apis/forecast-api-service.ts
```diff
@@ -41,6 +42,17 @@ export function CreateForecastApiService(): ForecastApiService {
             return response.json();
         },
 
+        /**
```
Jdallos: Please check indentation on all files, spaces should be set to 2 not 4 and at least in this file they are set to 2.

src/apis/schemas/time-entry.ts
```diff
@@ -4,4 +4,5 @@ interface TimeEntry {
     project: number,
     task: number,
     time_registered: number,
+    date: string,
```
Jdallos: All of these forecast related schemas should be grouped in a forecast folder, similar to how has been done for the pipedrive schemas. Of course this will require some updates to imports in various places.

src/functions/forecast/list-allocations/handler.ts
```diff
@@ -72,12 +72,29 @@ const listAllocations = async (api: ForecastApiService, currentDate: Date, param
 const listAllocationsHandler: ValidatedEventAPIGatewayProxyEvent<any> = async event => {
 
   const api = CreateForecastApiService();
+  let startDateParameter =  new Date();
+  let endDateParameter =  new Date();
+
+  if (event.queryStringParameters.startDate){
```
Jdallos: I believe in the case that there are no parameters passed queryStringParameters will not exist on the event. So there needs to be some error handling that queryStringParameters is defined before checking is startDate or endDate is defined, otherwise we will get a typeError for checking startDate on undefined.

You can see one way of handling this in the on-Call/list-on-call-handler.

Although perhaps abetter way here would be to use optional chaining in the  check e.g.  event?.queryStringParameters.startDate, this might work fine for this case and we can change it later if needed.

src/functions/forecast/list-allocations/handler.ts
```diff
@@ -72,12 +72,29 @@ const listAllocations = async (api: ForecastApiService, currentDate: Date, param
...
+    try {
+      startDateParameter = new Date(event.queryStringParameters.startDate);
+    } catch {
+      startDateParameter = new Date();
+    }
```
Jdallos: We only need to use try and catch when making asynchronous request so I see no reason for them here.
I think you could just use a ternary operator here. although perhaps I misunderstand what is happening here. For what reasons would you expect the catch to execute?

src/functions/forecast/list-allocations/handler.ts
```diff
@@ -72,12 +72,29 @@ const listAllocations = async (api: ForecastApiService, currentDate: Date, param
...
+  }
+  if (event.queryStringParameters.endDate){
+    try {
+      endDateParameter = new Date(event.queryStringParameters.endDate);
+    } catch {
+      endDateParameter = new Date();
+    }
```
Jdallos: Same issue here as above

src/functions/forecast/list-allocations/handler.ts
```diff
@@ -72,12 +72,29 @@ const listAllocations = async (api: ForecastApiService, currentDate: Date, param
...
+  let endDateParameter =  new Date();
+
+  if (event.queryStringParameters.startDate){
+    try {
+      startDateParameter = new Date(event.queryStringParameters.startDate);
+    } catch {
+      startDateParameter = new Date();
+    }
+  }
+  if (event.queryStringParameters.endDate){
```
Jdallos: Use the optional chaining as described in previous comment

src/functions/forecast/list-allocations/handler.ts
```diff
@@ -72,12 +72,29 @@ const listAllocations = async (api: ForecastApiService, currentDate: Date, param
...
+  }
 
   const allocations = await listAllocations(api, new Date(), {
-    startDate: new Date(event.queryStringParameters.startDate),
-    endDate: new Date(event.queryStringParameters.endDate),
-    personId: event.queryStringParameters.personId,
-    projectId: event.queryStringParameters.projectId,
+    startDate: startDateParameter,
+    endDate: endDateParameter,
+    personId: event.queryStringParameters.personId ? event.queryStringParameters.personId : undefined,
```
Jdallos: Slightly cleaner I think 
event.queryStringParameters.personId && event.queryStringParameters.personId
would work fine here.

Although perhaps you could destructure all parameters from the event at the start of this file and use the variable names directly rather than event.something.variable.

src/functions/forecast/list-allocations/handler.ts
```diff
@@ -72,12 +72,29 @@ const listAllocations = async (api: ForecastApiService, currentDate: Date, param
 const listAllocationsHandler: ValidatedEventAPIGatewayProxyEvent<any> = async event => {
 
   const api = CreateForecastApiService();
+  let startDateParameter =  new Date();
+  let endDateParameter =  new Date();
```
Jdallos: Any reason why these have parameter in the name? It seems they are not actually the "parameters" as passed to the lambda.

As described in a later comment I would recommend destructing the event.queryStringParameters at the top of this file and it might be worth renaming the startDate and endDate from the queryStringParameters as startDateParameter and endDateParameter (to avoid naming conflicts).

src/functions/forecast/list-allocations/handler.ts
```diff
@@ -72,12 +72,29 @@ const listAllocations = async (api: ForecastApiService, currentDate: Date, param
...
 
   const allocations = await listAllocations(api, new Date(), {
-    startDate: new Date(event.queryStringParameters.startDate),
-    endDate: new Date(event.queryStringParameters.endDate),
-    personId: event.queryStringParameters.personId,
-    projectId: event.queryStringParameters.projectId,
+    startDate: startDateParameter,
+    endDate: endDateParameter,
+    personId: event.queryStringParameters.personId ? event.queryStringParameters.personId : undefined,
+    projectId: event.queryStringParameters.projectId ? event.queryStringParameters.projectId : undefined,
```
Jdallos: Same as above, destructure variable from the queryStringParams and use the && syntax rather than ternary operator

src/functions/forecast/list-projects/handler.ts
```diff
@@ -10,7 +10,7 @@ import { CreateForecastApiService, ForecastApiService } from "src/apis/forecast-
 export interface ListProjectsParameters {
   startDate?: Date,
   endDate?: Date,
-  projectId?: string
+  projectId?: string,
```
Jdallos: No need for trailing comma here.

src/functions/forecast/list-projects/handler.ts
```diff
@@ -10,7 +10,7 @@ import { CreateForecastApiService, ForecastApiService } from "src/apis/forecast-
 export interface ListProjectsParameters {
```
Jdallos: Unless these interfaces are actually used in other files there is no need for them to be exported. So please remove the export for any of the Parameter and Response interfaces within the forecast handlers.

src/functions/forecast/list-projects/handler.ts
```diff
@@ -47,22 +48,74 @@ const listProjects = async (api: ForecastApiService, currentDate: Date, paramete
...
   const api = CreateForecastApiService();
+  let startDateParameter = new Date();
+  let endDateParameter = new Date();
 
+  if (event.queryStringParameters.projectId){
+    const project = await listProject(api, {projectId: event.queryStringParameters.projectId});
+    if (!project.id) {
+      return {
+        statusCode: 400,
+        body: "Invalid projectd"
```
Jdallos:
```suggestion
        body: "Invalid projectId"
```

src/functions/forecast/list-projects/handler.ts
```diff
@@ -47,22 +48,74 @@ const listProjects = async (api: ForecastApiService, currentDate: Date, paramete
...
  * Lambda for listing Forecast projects
  * 
  * @param event event
  */
 const listProjectsHandler: ValidatedEventAPIGatewayProxyEvent<any> = async event => {
   const api = CreateForecastApiService();
+  let startDateParameter = new Date();
+  let endDateParameter = new Date();
 
+  if (event.queryStringParameters.projectId){
```
Jdallos: Similar issues here as described in the allocations handler, use optional chaining for the event?.queryStringParamerters check.
Destructure the queryStringParams, rename the startDateParameter and endDateParameter variables as I suggested (or handle some other way as you see fit).

src/functions/forecast/list-projects/handler.ts
```diff
@@ -47,22 +48,74 @@ const listProjects = async (api: ForecastApiService, currentDate: Date, paramete
...
+    const project = await listProject(api, {projectId: event.queryStringParameters.projectId});
+    if (!project.id) {
+      return {
+        statusCode: 400,
```
Jdallos: I think 204 might be a more appropriate code in this case. Have a look and see what you think

src/functions/forecast/list-projects/handler.ts
```diff
@@ -47,22 +48,74 @@ const listProjects = async (api: ForecastApiService, currentDate: Date, paramete
...
+      startDateParameter = new Date();
+    }
+  }
+  if (event.queryStringParameters.endDate){
+    try {
+      endDateParameter = new Date(event.queryStringParameters.endDate);
+    } catch {
+      endDateParameter = new Date();
+    }
+  }
```
Jdallos: As mentioned in allocations handler, we need optional chaining here to be typesafe and I'm not sure why you are using try catch here. Lets discuss what conditions you would expect the catch blocks to execute and go from there.

src/functions/forecast/list-projects/handler.ts
```diff
@@ -47,22 +48,74 @@ const listProjects = async (api: ForecastApiService, currentDate: Date, paramete
...
   const filteredProjects = await listProjects(api, new Date(), {
-    startDate: new Date(event.queryStringParameters.startDate),
-    endDate: new Date(event.queryStringParameters.endDate),
+    startDate: startDateParameter,
+    endDate: endDateParameter,
```
Jdallos: No need for a trailing comma on the final line here

src/functions/forecast/list-tasks/handler.ts
```diff
@@ -68,7 +68,7 @@ const listTasksHandler: ValidatedEventAPIGatewayProxyEvent<any> = async event =>
   const api = CreateForecastApiService();
 
   const tasks = await listTasks(api, {
-    projectId: parseInt(event.queryStringParameters.projectId),
+    projectId: event.queryStringParameters ? parseInt(event.queryStringParameters.projectId) : undefined,
```
Jdallos: && syntax here should be fine

src/functions/forecast/list-projects/handler.ts
```diff
@@ -10,7 +10,7 @@ import { CreateForecastApiService, ForecastApiService } from "src/apis/forecast-
 export interface ListProjectsParameters {
   startDate?: Date,
   endDate?: Date,
-  projectId?: string
+  projectId?: string,
 }
 
 /**
```
Jdallos, line 15: There is an unused commented out import at the top of this file, please remove

src/functions/forecast/list-time-entries/handler.ts
```diff
@@ -29,14 +34,21 @@ export interface Response {
...
+  });
+
+  return filteredTimeEntries.map(timeEntry => {
     return {
       id: timeEntry.id,
       person: timeEntry.person,
       project: timeEntry.project,
       task: timeEntry.task,
       timeRegistered: timeEntry.time_registered,
+      date: timeEntry.date,
```
Jdallos: No need for a comma on the last line as in this case.

src/functions/forecast/list-time-entries/handler.ts
```diff
@@ -58,6 +70,9 @@ const listTimeEntriesHandler: ValidatedEventAPIGatewayProxyEvent<any> = async ev
   
   const timeEntries = await listTimeEntries(api, {
     projectId: parseInt(event.queryStringParameters.projectId),
+    startDate: event.queryStringParameters.startDate ? event.queryStringParameters.startDate : undefined,
```
Jdallos: I can't leave a comment on the appropriate line due to github limitations but on the first line of the listTimeEntriesHandler we need the optional chaining for !event?.queryStringParameters.projectId, perhaps the message could also be updated to Missing parameters rather than Invalid parameters

src/functions/forecast/list-time-entries/handler.ts
```diff
@@ -58,6 +70,9 @@ const listTimeEntriesHandler: ValidatedEventAPIGatewayProxyEvent<any> = async ev
...
+    endDate: event.queryStringParameters.endDate ? event.queryStringParameters.endDate : undefined,
+    taskId: event.queryStringParameters.taskId ? event.queryStringParameters.taskId : undefined,
```
Jdallos: Would && syntax be cleaner here rather than ternary operator

src/functions/forecast/list-time-entries/handler.ts
```diff
@@ -29,14 +34,21 @@ export interface Response {
  */
 const listTimeEntries = async (api: ForecastApiService, parameters: ListTimeEntriesParameters): Promise<Response[]> => {
   const timeEntries = await api.getTimeEntriesByProject(parameters.projectId);
+  const currentDate = new Date();
 
-  return timeEntries.map(timeEntry => {
+  const filteredTimeEntries = timeEntries.filter(timeEntry => {
+    return FilterUtilities.filterByDate({start_date: parameters.startDate, end_date: parameters.endDate}, currentDate, {startDate: new Date(timeEntry.date), endDate: new Date(timeEntry.date)})
+    && FilterUtilities.filterByTask(timeEntry.task, parameters.taskId);
```
Jdallos: Use indentation to make this block a little more readable.

src/libs/filter-utils.ts
```diff
@@ -61,4 +61,16 @@ export namespace FilterUtilities {
...
   }
+
+  /**
+   * Compares Forecast task id to specified id
+   * 
+   * @param task Task id from Forecast
+   * @param taskId Task id to compare
+   * @returns If two parameters match or task is null
+   */
+  export const filterByTask = (task?: number, taskId?: string): boolean => {
```
Jdallos: No need for the explicit return typing here.

Jdallos (changes requested):
Changes requested

src/apis/forecast-api-service.ts
```diff
@@ -1,103 +1,103 @@
...
 
 export interface ForecastApiService {
-    getAllocations: () => Promise<Allocation[]>;
-    getProjects: () => Promise<Project[]>;
-    getProject: (projectId: string) => Promise<Project>;
-    getTasksByProject: (projectId: number) => Promise<Task[]>;
-    getAllTasks: () => Promise<Task[]>;
-    getProjectSprints: (projectId: number) => Promise<Sprint[]>;
-    getTimeEntriesByProject: (projectId: number) => Promise<TimeEntry[]>;
+	getAllocations: () => Promise<Allocation[]>;
```
Jdallos, line 7: Spaces are set to 4 they should be 2 as said in a previous comment

Jdallos: Actually this seems fine in the VSCode, perhaps its just the github UI not showing indentation correctly. If it seems to you the indentation is set to 2 spaces then this is fine.

src/functions/forecast/list-time-entries/handler.ts
```diff
@@ -37,8 +37,11 @@ const listTimeEntries = async (api: ForecastApiService, parameters: ListTimeEntr
   const currentDate = new Date();
 
   const filteredTimeEntries = timeEntries.filter(timeEntry => {
-    return FilterUtilities.filterByDate({start_date: parameters.startDate, end_date: parameters.endDate}, currentDate, {startDate: new Date(timeEntry.date), endDate: new Date(timeEntry.date)})
-    && FilterUtilities.filterByTask(timeEntry.task, parameters.taskId);
+    return FilterUtilities.filterByDate(
+      {start_date: parameters.startDate, end_date: parameters.endDate}, 
+      currentDate, 
+      {startDate: new Date(timeEntry.date), endDate: new Date(timeEntry.date)}
```
Jdallos:
```suggestion
        {start_date: parameters.startDate, end_date: parameters.endDate}, 
        currentDate, 
        {startDate: new Date(timeEntry.date), endDate: new Date(timeEntry.date)}
```

Jdallos (approved):
Couple of comments and approved

PR #34, opened 2024-03-25
Add slack avatar image handler

### LINKED ISSUE #32: Get user avatars from Slack

Kseniia-Koshkina:


DZotoff (approved):

src/functions/meta-assistant/slack-user-avatar/handler.ts
```diff
@@ -0,0 +1,50 @@
+import { ValidatedEventAPIGatewayProxyEvent } from "@libs/api-gateway";
+import { middyfy } from "@libs/lambda";
+import SlackUtilities from "src/meta-assistant/slack/slack-utils";
+
+/**
+ * Parameters for lambda
+ */
+export interface SlackUsersAvatarsParameters {
```
Jdallos: Does this need to be exported

src/functions/meta-assistant/slack-user-avatar/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+  email: string,
+}
+
+/**
+ * Response schema for lambda
+ */
+export interface Response {
+  image_original: string,
+	image_24: string,
+	image_32: string,
```
Jdallos: Do we need to return 3 different types of image?

src/functions/meta-assistant/slack-user-avatar/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
```
Jdallos: I have reviewed the spec and suggested changing this to return  all slack users images in a single response rather than us having to request this for every user. Let me know if you think this idea would work or if there are some problems with it.

Jdallos (changes requested):

DZotoff (approved):

src/functions/forecast/list-tasks/mockData.json
Jdallos, line 1: We don't want to commit mockData

src/functions/slack-user-avatar/handler.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (!accessToken) {
+    throw new Error("Timebank authentication failed");
+  }
+
+  const timebankUsers = await TimeBankApiProvider.getTimebankUsers(accessToken);
+  if (!timebankUsers) {
+    throw new Error("No persons retrieved from Timebank");
+  }
+
+  const slackUsers = await SlackUtilities.getSlackUsers();
```
Jdallos: It's probably a good idea to wrap these requests in a try catch and handle the errors/ error messages appropriately e.g. w´there is no error handling currently for if the getSlackUsers request fails

src/functions/slack-user-avatar/mockData.json
Jdallos, line 1: No need to commit mockData.json

Jdallos (approved):
Some changes requested and approved

PR #37, opened 2024-04-16
MetaAssistant holidays bug

### LINKED ISSUE #36: MetaAssistant holidays bug

Kseniia-Koshkina:


src/meta-assistant/slack/slack-utils.ts
```diff
@@ -149,6 +149,18 @@ Have a great week!
...
   );
 
+  /**
+   * Check if the person was not supposed to work and there is no record of his activity.
+   * 
+   * @param expected expected number of minutes
+   * @param logged logged minutes
+   * @returns true if user's expected and logged were 0
+   */
+  const checkIfHoliday = (expected: number, logged: number) => {
```
Jdallos, line 159: There should be some tests created to clearly demonstrate this works as intended. 
As a reference a message was sent on 2nd May referring to 1st May, if this works as intended then the message should not be sent as the 1st is a national holiday

Jdallos (changes requested):

PR #50, opened 2024-08-22
convert allocations to Severa

### LINKED ISSUE #49: convert to Severa usage

Essomatic:
this is here just to save progress since I'm leaving soon. We should consider what we want from the allocations response from severa and add response types accordingly.  add userGuids or projectGuids to mockData.json at list-allocations/mockData.json to test the function. this same endpoint could be used to search for user(s) or project(s) allocations or the intersection (A ∩ B) 

Command to test: npx sls invoke local -f listAllocationsHandler --stage dev --path src/functions/severa/list-allocations/mockData.json


package-lock.json
Essomatic, line 1: what is going on over here

PR #68, opened 2024-09-19
Feature 63 create find lambdas for questionnaire

### LINKED ISSUE #63: Create & find lambdas for Questionnaire
functions for questionnaire, making lambdas for create quiz and finding quiz.

dalju005-xamk:


serverless.ts
```diff
@@ -36,7 +38,7 @@ const serverlessConfiguration: AWS = {
   provider: {
     name: 'aws',
     runtime: 'nodejs16.x',
-    region: env.AWS_DEFAULT_REGION as any,
+    region: (env.AWS_DEFAULT_REGION as any) || "us-east-1",
```
McDuuck, line 41: This not sure if it needs to be env or hardcoded region.

src/functions/questionnaire/create-quiz/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    let data: QuestionnaireModel;
+    if (typeof event.body === "string") {
+      data = JSON.parse(event.body);
+    } else {
+      data = event.body as QuestionnaireModel;
+    }
+
+    if (!data.id || !data.title) {
```
McDuuck: pretty sure there is more required options than these two

serverless.ts
Jdallos: This is fine

serverless.ts
```diff
@@ -28,6 +28,8 @@ import updateSoftwareHandler from "@/functions/software-registry/update-software
 import deleteSoftwareHandler from "@/functions/software-registry/delete-software";
 import listUsersHandler from "@/functions/keycloak/list-users";
 import findUserHandler from "src/functions/keycloak/find-user";
+import createQuizHandler from "src/functions/questionnaire/create-quiz";
```
Jdallos: We should be consistent in the naming, we are using both quiz and questionnaire interchangably, pick one. My preference is questionnaire but your choice

src/functions/questionnaire/create-quiz/handler.ts
```diff
@@ -0,0 +1,66 @@
+import QuestionnaireService from "src/database/services/quiz-api-service";
+import { middyfy } from "src/libs/lambda";
+import { DocumentClient } from "aws-sdk/clients/dynamodb";
+import type { QuestionnaireModel } from "src/database/schemas/questionnaire/questionnaire";
+import type { ValidatedEventAPIGatewayProxyEvent } from "src/libs/api-gateway";
+
+const dynamoDb = new DocumentClient();
+const questionnaireService = new QuestionnaireService(dynamoDb);
```
Jdallos: This can be removed and you can import the service e.g.
import { questionnaireService } from "src/database/services";

As it is exported from the database/serveices/index.ts

src/functions/questionnaire/create-quiz/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Handler for creating a new questionnaire entry in DynamoDB.
+ *
+ * @param event - API Gateway event containing the request body.
+ * @returns Response object with status code and body.
+ */
+export const createQuizHandler: ValidatedEventAPIGatewayProxyEvent<QuestionnaireModel> = async (event) => {
```
Jdallos: It might be that we are using the DB model here when we should be using the schema. For reference see in elastic app repos the src/schema/curation vs the src/database/models/curation. Also note how the schema is used in the createCuration handler as the type for the lambda.

I thin the main reason for this is that it allows us to have some differences between the DB type of entity and the body sent to the lambda describing the entity to be created. E.g. id can be created in the lambda rather than form the UI

src/functions/questionnaire/create-quiz/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ error: "Request body is required." })
+      };
+    }
+
+    let data: QuestionnaireModel;
+    if (typeof event.body === "string") {
+      data = JSON.parse(event.body);
+    } else {
```
Jdallos: I don't think this should be required, as the body should not be a string.

src/functions/questionnaire/create-quiz/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+      data = event.body as QuestionnaireModel;
```
Jdallos: Hopefully we don't need to cast here as type can be inferred from the function definition.
Also would recommend destructuring all attributes from the body as done in createCuration function.

src/functions/questionnaire/create-quiz/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    if (!data) {
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ error: "Some required data is missing !" })
+      };
+    }
```
Jdallos: This could be updated to check for all of the required attributes from the body- once they are destructured

src/functions/questionnaire/create-quiz/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+      id: data.id,
+      title: data.title,
+      description: data.description,
+      options: data.options,
+      tags: data.tags,
+      passedUsers: data.passedUsers,
+      passScore: data.passScore
+    };
+
+    const createdQuestionnaire = await questionnaireService.createQuestionnaire(newQuestionnaire);
```
Jdallos: These could be combined, here is example form elastic app functions, create curation:
      const createdDocument = await documentService.createDocument({
        id: newDocumentId,
        title: title,
        description: description,
        links: links,
        language: language,
        curationId: curationId
      });

src/functions/questionnaire/create-quiz/index.ts
```diff
@@ -0,0 +1,13 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "post",
+        path: "/admin/questionnaire"
```
Jdallos:
```suggestion
        path: "/questionnaire"
```

If we are considering that only admins in the UI can create the questionnaire then we don't need to change the endpoint. We might want to do some investigation into making sure the requesting user has the correct role. Lets add a comment about checking the role in a future feature as I'm not sure about the best way to approach this right now. Probably some kind of function to decode the token and check use has the role of "admin". Please add a task with this description to the trello backlog column (at the top)

src/functions/questionnaire/find-quiz/handler.ts
```diff
@@ -0,0 +1,60 @@
+import QuestionnaireService from "src/database/services/quiz-api-service";
+// import { CreateKeycloakApiService, type KeycloakApiService } from "src/apis/keycloak-api-service";
```
Jdallos: Remove commented code

src/functions/questionnaire/find-quiz/handler.ts
```diff
@@ -0,0 +1,60 @@
...
+import { middyfy } from "src/libs/lambda";
+import { DocumentClient } from "aws-sdk/clients/dynamodb";
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+
+const dynamoDb = new DocumentClient();
+const questionnaireService = new QuestionnaireService(dynamoDb);
```
Jdallos: These can be removed as described in the create quiz handler

src/functions/questionnaire/find-quiz/handler.ts
```diff
@@ -0,0 +1,60 @@
...
+  event: APIGatewayProxyEvent,
+) => {
+  const { id } = event.pathParameters || {};
+
+  try {
+    if (!id) {
+      return {
+        headers: {
+          "Content-Type": "application/json",
+        },
```
Jdallos:
```suggestion
```

src/functions/questionnaire/find-quiz/handler.ts
```diff
@@ -0,0 +1,60 @@
...
+        statusCode: 400,
+        body: JSON.stringify({
+          error: "Missing or invalid path parameter: id",
+        }),
+      };
+    }
+
+    const quizById = await questionnaireService.findQuestionnaire(id);
+
+    if (!quizById) throw new Error("Quiz not found");
```
Jdallos: Instead of throw just return as in other places, code 404 and error message you have is fine "Quiz not found"

src/functions/questionnaire/find-quiz/handler.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    return {
+      headers: {
+        "Content-Type": "application/json",
+      },
```
Jdallos:
```suggestion
```

src/functions/questionnaire/find-quiz/handler.ts
```diff
@@ -0,0 +1,60 @@
...
+      statusCode: 200,
+      body: JSON.stringify(quizById),
+    };
+  } catch (error) {
+    return {
+      headers: {
+        "Content-Type": "application/json",
+      },
```
Jdallos:
```suggestion
```

src/functions/questionnaire/find-quiz/handler.ts
Jdallos: I'm quite sure you don't need to return headers here and you have not done this in the create function

src/functions/questionnaire/find-quiz/index.ts
```diff
@@ -0,0 +1,13 @@
+import { handlerPath } from '@libs/handler-resolver';
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "get",
+        path: "/questionnaire/{id}",
```
Jdallos:
```suggestion
        path: "/questionnaire/{id}",
        authorizer: {
          name: "timebankKeycloakAuthorizer"
        }
```

Jdallos: Need to add the authorizer as this is what allows the logged in user in metatavu home to be authorized based on the keycloak access token.

src/functions/questionnaire/create-quiz/index.ts
Jdallos:
```suggestion
        path: "/questionnaire",
          authorizer: {
          name: "timebankKeycloakAuthorizer"
        }
```

Jdallos: Need to add the authorizer as this is what allows the logged in user in metatavu home to be authorized based on the keycloak access token.

Jdallos (changes requested):

src/functions/questionnaire/create-quiz/handler.ts
dalju005-xamk: This should be build. Now its different in our code than it is in elastic

dalju005-xamk: Now it its included to this PR. Please check those also.

src/functions/questionnaire/create-questionnaire/handler.ts
```diff
@@ -0,0 +1,58 @@
...
+  const { title, description, options, tags, passedUsers, passScore } = JSON.parse(event.body);
+  
+  if (!title || !description || !options || !passScore) {
+    return {
+      statusCode: 400,
+      body: JSON.stringify({ error: "Some required data is missing !" })
+    };
+  }
+
+  const newQuestionnaireId: string | undefined = uuidv4();
```
Jdallos: Does this need to be | undefined?

src/functions/questionnaire/find-questionnaire/handler.ts
```diff
@@ -14,17 +9,12 @@ const questionnaireService = new QuestionnaireService(dynamoDb);
  * @returns quiz information as object
  */
 
-const findQuizHandler: APIGatewayProxyHandler = async (
-  event: APIGatewayProxyEvent,
-) => {
+const findQuestionHandler: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {
```
Jdallos:
```suggestion
const findQuestionnaireHandler: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {
```

src/functions/questionnaire/create-questionnaire/handler.ts
```diff
@@ -0,0 +1,58 @@
...
+import type QuestionnaireModel from "src/database/models/questionnaire";
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+
+/**
+ * Handler for creating a new questionnaire entry in DynamoDB.
+ *
+ * @param event - API Gateway event containing the request body.
+ * @returns Response object with status code
+ */
+export const createQuestionHandler: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {
```
Jdallos:
```suggestion
export const createQuestionnaireHandler: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {
```

src/schema/questionnaire.ts
```diff
@@ -0,0 +1,25 @@
+import { Type } from "@sinclair/typebox";
+/**
+ * Schema for the Questionnaire table
+ */
+
```
Jdallos:
```suggestion

/**
 * Schema for the Questionnaire table
 */
```

src/schema/questionnaire.ts
```diff
@@ -0,0 +1,25 @@
...
+      options: Type.Array(
+        Type.Object({
+          label: Type.String(),
+          value: Type.Boolean(),
+        }),
+      ),
+    }),
+  ),
+  tags: Type.Optional(Type.Array(Type.String())),
+  passedUsers: Type.Optional(Type.Array(Type.Number())),
```
Jdallos, line 21: This schema seems fine, however I notice only tags and passed users are optional (also fine) but in the home-lambdas-api swagger the questionnaire schema does not have any required properties. This could cause issue down the line so I would suggest creating a task in trello for updating the spec to contain the appropriate required values. This feature can be merged (and even tested) before the spec is updated but I would not recommend doing any UI work using the lambdas until the spec is updated.

Jdallos (approved):
Some changes

src/schema/questionnaire.ts
dalju005-xamk: Made new Trellocard for this. Added parts of this comment to that.

PR #70, opened 2024-09-23
Update lambda function

### LINKED ISSUE #67: Update lambda function for questionnaire

McDuuck:


src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,49 @@
+import { ValidatedEventAPIGatewayProxyEvent } from "@libs/api-gateway";
```
dalju005-xamk: Perhaps use type of here ?
```suggestion
import type { ValidatedEventAPIGatewayProxyEvent } from "@libs/api-gateway";
```

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,49 @@
...
+import { middyfy } from "@libs/lambda";
+import { questionnaireService } from "src/database/services";
+import QuestionnaireModel from "src/database/schemas/questionnaire/questionnaire";
+
+/**
+ * Lambda function to update a questionnaire
+ *
+ * @param event event
+ */
+const updateQuestionnaire: ValidatedEventAPIGatewayProxyEvent<typeof QuestionnaireModel> = async event => {
```
dalju005-xamk: Perhaps like this ?
```suggestion
const updateQuestionnaire: ValidatedEventAPIGatewayProxyEvent<typeof QuestionnaireModel> = async event => 
```

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const existingQuestionnaire = await questionnaireService.findQuestionnaire(id);
+  if (!existingQuestionnaire) {
+    return {
+      statusCode: 404,
+      body: `Questionnaire ${id} not found`
+    };
+  }
+
+  const questionnaireUpdates: QuestionnaireModel = {
```
dalju005-xamk, line 39: Make Import already. The model is now in develop.

```suggestion
  import type QuestionnaireModel from "src/database/models/questionnaire";
```

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,49 @@
...
+    ...existingQuestionnaire,
+    ...body
+  };
+
+  try {
+    const updatedQuestionnaire = await questionnaireService.updateQuestionnaire(questionnaireUpdates);
```
dalju005-xamk, line 50: Import this also.

```suggestion
   import { questionnaireService } from "src/database/services";
```

AfoninOleksandr (approved):

dalju005-xamk (approved):

serverless.ts
Jdallos, line 1: This branch needs to be merged with the updated develop, you are missing updated table name in serverless.ts and fetch import for the keycloak service.
Also please check this is the serverless.ts:
AttributeDefinitions: [{ AttributeName: "id", AttributeType: "N" }],

I've just approved a spec change changing questionnaire id to be a string, so this may need updating

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+import { questionnaireService } from "src/database/services";
+import type QuestionnaireModel from "src/database/models/questionnaire";
+import type questionnaireSchema from "src/schema/questionnaire";
+
+/**
+ * Lambda function to update a questionnaire
+ *
+ * @param event event
+ */
+const updateQuestionnaire: ValidatedEventAPIGatewayProxyEvent<typeof questionnaireSchema> = async event => {
```
Jdallos: Other functions are named createQuestionnaire**Handler** etc. Lets keep the naming consistent so lets rename this updateQuestionnaireHandler

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+  const { pathParameters, body, } = event;
+  const id = pathParameters?.id;
+
+  if (!id) {
+    return {
+      statusCode: 400,
+      body: "Bad request"
```
Jdallos:
```suggestion
      body: "Bad request, missing id"
```

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+import type QuestionnaireModel from "src/database/models/questionnaire";
+import type questionnaireSchema from "src/schema/questionnaire";
+
+/**
+ * Lambda function to update a questionnaire
+ *
+ * @param event event
+ */
+const updateQuestionnaire: ValidatedEventAPIGatewayProxyEvent<typeof questionnaireSchema> = async event => {
+  const { pathParameters, body, } = event;
```
Jdallos:
```suggestion
  const { pathParameters, body } = event;
```

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+    return {
+      statusCode: 404,
+      body: `Questionnaire ${id} not found`
+    };
+  }
+
+  const questionnaireUpdates: QuestionnaireModel = {
+    ...existingQuestionnaire,
+    ...body
+  };
```
Jdallos: We should allocate to each attribute specifically here. The UI will fetch the existing questionnaire to be updated, the user will update the given fields, and then the whole questionnaire will be sent to the lambda as is. In this case probably only id needs to come from the existingQuestionnaire (although you have that from the lambda event as id also, i'd still use if from existingQuestionnaire).

Here is an example of what I mean from updateCurations:
  const curationUpdates: CurationModel = {
    id: existingCuration.id,
    name: name,
    elasticCurationId: existingCuration.elasticCurationId,
    documentId: existingCuration.documentId,
    queries: queries,
    promoted: curationType === CurationType.CUSTOM ? existingCuration.promoted : promoted,
    hidden: curationType === CurationType.CUSTOM ? existingCuration.hidden : hidden,
    startTime: startTime,
    endTime: endTime,
    curationType: curationType,
    groupId: groupId,
    language: language,
    invalidDocuments: invalidDocuments
  };
  
  In this case the attributes e.g. name, queries etc have already been destructured from the body, you should do similar

Jdallos (changes requested):
Changes requested

serverless.ts
Jdallos: It's still AttributeDefinitions: [{ AttributeName: "id", AttributeType: "N" }],


But I thought it should not be S as the spec was changed yesterday?

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  const questionnaireUpdates: QuestionnaireModel = {
+    id: existingQuestionnaire.id,
+    title: existingQuestionnaire.title,
+    description: existingQuestionnaire.description,
+    options: existingQuestionnaire.options,
+    tags: existingQuestionnaire.tags,
+    passedUsers: existingQuestionnaire.passedUsers,
+    passScore: existingQuestionnaire.passScore,
```
Jdallos: The only thing that should be coming form the existing questionnaire is the id. All the other attributes should come from the body, otherwise we are not actually updating the questionnaire.
As mentioned in a comment previously I would suggest destructuring the properties from the event body at the start of the lambda once you have destructured the body from the event.

e.g.
  const { title, description, options, tags, passedUsers, passScore } = body;
  
  See the elastic update curations as a reference

Jdallos (changes requested):

src/functions/questionnaire/update-questionnaire/handler.ts
```diff
@@ -12,6 +12,15 @@ import type questionnaireSchema from "src/schema/questionnaire";
...
   const id = pathParameters?.id;
+  const {
+    title,
+    description,
+    options,
+    tags,
+    passedUsers,
+    passScore,
+  } = body;
+
```
Jdallos, line 23:
```suggestion
```

Jdallos (approved):

PR #85, opened 2024-10-03
Feature 84 lambda for severaapi

### LINKED ISSUE #84: Lambda for SeveraAPI
Create Lambda for getting flextime from Severa.

dalju005-xamk:
getFlexTime Lambda for getting flextimes from Severa


McDuuck (approved):

src/database/models/severa.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * FlextimeModel for SeveraApiService.
+ */
+interface FlextimeModel {
+	totalFlextimeBalance: number | null; 
+  monthFlextimeBalance: number | null;
+}
+
+export default FlextimeModel;
```
Jdallos: Fix indentation, is there any reason the type is set as or null?

Also this is not a model as we are not persisting this in a database. 

We need to fix the contents under database/. Please move the forecast, keycloak and pipedrive folders and contents under database/schemas to src/types.
Then move database/schemas/software-regsitry into models. At that point database/schemas should be empty and can be deleted.

Jdallos: Please ensure all imports are updated as you do this.

src/database/services/severa-api-service.ts
```diff
@@ -0,0 +1,76 @@
+import fetch from "node-fetch";
+import type FlextimeModel from "../models/severa";
+
+/**
+ * Creates SeveraApiService
+ */
+export const CreateSeveraApiService = () => {
```
Jdallos: This is likely to be expanded so lets create an interface to describe the return value (see forecasreApiService).

src/database/services/severa-api-service.ts
```diff
@@ -0,0 +1,76 @@
...
+  const baseUrl: string = process.env.SEVERA_BASE_URL;
+
+  return {
+    /**
+     * Gets flextime by userGUID and eventDate
+     */
+    getFlextimeByUser: async (
+      userGuid: string,
+      eventDate: string,
```
Jdallos: I suggest removing this in spec PR

src/database/services/severa-api-service.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+export const CreateSeveraApiService = () => {
+  const baseUrl: string = process.env.SEVERA_BASE_URL;
+
+  return {
+    /**
+     * Gets flextime by userGUID and eventDate
+     */
+    getFlextimeByUser: async (
+      userGuid: string,
```
Jdallos: I suggest renaming all userGuid references to severaGuid as it is names in the users schema in the spec.

src/database/services/severa-api-service.ts
```diff
@@ -0,0 +1,76 @@
...
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(requestBody),
+    });
+
+    if (!response.ok) {
+      throw new Error(
+        `Failed to get access token: ${response.status} - ${response.statusText}`,
```
Jdallos:
```suggestion
        `Failed to get Severa access token: ${response.status} - ${response.statusText}`,
```

src/database/services/severa-api-service.ts
```diff
@@ -0,0 +1,76 @@
...
+      );
+    }
+    const data = await response.json();
+
+    return data.access_token;
+  } catch (error) {
+    throw new Error(`Failed to get access token: ${error.message}`);
```
Jdallos:
```suggestion
    throw new Error(`Failed to get Severa access token: ${error.message}`);
```

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -0,0 +1,38 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+import type FlextimeModel from "src/database/models/severa";
+import { CreateSeveraApiService } from "src/database/services/severa-api-service";
+
+/**
+ * @param event
+ * @returns users flextime
+ */
+export const getFlextimeHandler: APIGatewayProxyHandler = async (
```
Jdallos: It seems for some functions we set the return type as APIGatewayProxyHandler and for others we set it as ValidatedEventAPIGatewayProxyEvent<any>. Please investigate the two and then change this accordingly.

It looks to me like ValidatedEventAPIGatewayProxyEvent<any> is the way it's done in Elastic-app-functions which should be our example.
Also in most cases the any can be replaced with the typeof schema- see schame/questionnaire.ts for an example. Might be worth discussing this with Ilmo as to whether the questionnaire lambdas should also be updated.

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -0,0 +1,38 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+import type FlextimeModel from "src/database/models/severa";
+import { CreateSeveraApiService } from "src/database/services/severa-api-service";
+
+/**
+ * @param event
+ * @returns users flextime
```
Jdallos:
```suggestion
```

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -0,0 +1,38 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+import type FlextimeModel from "src/database/models/severa";
+import { CreateSeveraApiService } from "src/database/services/severa-api-service";
+
+/**
+ * @param event
```
Jdallos: Missing description

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @returns users flextime
+ */
+export const getFlextimeHandler: APIGatewayProxyHandler = async (
+  event: APIGatewayProxyEvent,
+) => {
+  try {
+    const userGuid = event.pathParameters?.userGuid;
```
Jdallos: rename userGuid as severaGuid

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -0,0 +1,38 @@
...
+    const eventDate = event.queryStringParameters?.eventDate;
```
Jdallos: As commented in the spec PR we should set the date to be the current date for now, we just want to know todays/ this months balance.

Jdallos: Meaning eventDate can be removed- we should check what the default value is set as by the severa API as it is not a required param in their docs

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -0,0 +1,38 @@
...
+    };
+  } catch (error) {
+    return {
+      statusCode: 500,
+      body: JSON.stringify({ error: error.message }),
+    };
+  }
+};
+
+export const main = getFlextimeHandler;
```
Jdallos: Missing the middyfy

src/functions/severa/get-flextime-by-user/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from '@libs/handler-resolver';
+
+export default {
+	handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "get",
+        path: "/v1/users/{userGuid}/flextime",
```
Jdallos: Update as commented in the spec

src/functions/severa/get-flextime-by-user/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from '@libs/handler-resolver';
```
Jdallos: Double quotes

src/functions/severa/get-flextime-by-user/handler.ts
Jdallos: Also needs to be exported from the functions/index.ts

There should be no problem testing this by invoking it locally so make sure this is done- I suspect it would not actually invoke without the above changes.

Jdallos (changes requested):

src/database/models/severa.ts
dalju005-xamk: These are now done. I made changes in own commit.

src/database/services/severa-api-service.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      const response = await fetch(url, {
+        method: "GET",
+        headers: {
+          Authorization: `Bearer ${await getSeveraAccessToken()}`,
+          "Client_Id": process.env.SEVERA_DEMO_CLIENT_ID,
+          "Content-Type": "application/json",
+        },
+      });
+
```
naStipe, line 32: Would it be a good idea to cover it with try-catch block?

src/functions/index.ts
```diff
@@ -28,4 +28,5 @@ export { default as findQuestionnaireHandler } from "./questionnaire/find-questi
 export { default as deleteQuestionnaireHandler } from "./questionnaire/delete-questionnaire";
 export { default as listQuestionnaireHandler } from "./questionnaire/list-questionnaire";   
 export { default as updateQuestionnaireHandler } from "./questionnaire/update-questionnaire";
+export { default as getFlextimeHandler } from "./severa/get-flextime-by-user";
```
naStipe, line 31: I would maybe create additional directory for flextime, like "./severa/flextime/get-flextime-by-user" to later add directories for Allocations and so on

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (!severaGuid) {
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ message: "User Guid is required" }),
+      };
+    }
+
+    const eventDate = eventDateParam || new Date().toISOString();
+
+    const flexTimeBySeveraGuid = await api.getFlextimeBySeveraGuid(severaGuid, eventDate);
```
naStipe: Do we actually need to convert the date if we are just gonna put the same date back in the string(path)?

src/functions/severa/get-flextime-by-user/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+	handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "get",
+        path: "/severa/flextime/{severaGuid}",
```
naStipe: I would suggest to not use severaGuid and use severaUserGuid and severaProjectGuid. Although, I am not sure if it is not gonna get messy or not.

naStipe (approved):
Looks good, but I noted parts that raised minor questions for me at least.

Hanhz2707 (approved):

Jdallos (approved):

PR #90, opened 2024-10-17
Bugfix 87 fixing lambdas for keycloak

### LINKED ISSUE #87: Fixing lambdas for Keycloak

dalju005-xamk:
Fixing both lambdas to return same type of information. Small changes to api also.
User model is in new path, because folder structure is going to change in PR: feature-84


naStipe (approved):

src/database/services/keycloak-api-service.ts
```diff
@@ -1,12 +1,12 @@
...
 import fetch from "node-fetch";
 
 /**
  * Interface for a KeycloakApiService.
  */
 export interface KeycloakApiService {
   getUsers: () => Promise<User[]>;
-  findUser: (id: string) => Promise<User[]>;
+  findUser: (id: string) => Promise<User>;
 }
```
Hanhz2707, line 11: Maybe we can won't need this empty space as Josh reviewed our old PR and commented about this

Hanhz2707 (approved):

src/functions/keycloak/find-user/handler.ts
```diff
@@ -1,19 +1,10 @@
 import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
 import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
 import { middyfy } from "src/libs/lambda";
-
+import type { User } from "src/types/keycloak/user";
 /**
- * Response schema for lambda
+ * FIXME: At this moment in KeyCloak its called: severa-user-id (string), in here its severaGuid (string);
```
Jdallos: If this relates to the severaGuid attribute in filteredUser then I'd suggest putting the comment directly above the attribute. However if the attribute name is defined in the User type, then I'd suggest putting this comment there- unless there is a reason not to?

src/functions/keycloak/find-user/handler.ts
```diff
@@ -33,11 +24,27 @@ const findUserHandler: APIGatewayProxyHandler = async (
       throw new Error("Missing or invalid path parameter: id");
     }
 
-    const userById = await api.findUser(queryStringParameters.id);
+    const userById: User = await api.findUser(queryStringParameters.id);
```
Jdallos: I don't think this needs explicitly typing as User

src/functions/keycloak/find-user/handler.ts
```diff
@@ -33,11 +24,27 @@ const findUserHandler: APIGatewayProxyHandler = async (
...
+    }
+    const filteredUser: User = {
+      id: userById.id,
+      firstName: userById.firstName,
+      lastName: userById.lastName,
+      email: userById.email,
+      isActive: userById.isActive,
+      severaGuid: userById.severaGuid,
+      forecastId: userById.forecastId
+    };
```
Jdallos: I'm a bit confused, why are we filtering user here? Is it because the response from the api.findUser is not actually of User type? In which case the response type should be updated to reflect what is actually returned from the request.

src/database/services/keycloak-api-service.ts
```diff
@@ -45,7 +45,7 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
      * @param id string
      * @returns user by Id
      */
-    findUser: async (id: string): Promise<User[]> => {
+    findUser: async (id: string): Promise<User> => {
```
Jdallos: Is this actually returning something of type User?

src/functions/keycloak/list-users/handler.ts
```diff
@@ -1,22 +1,10 @@
...
-import {
-  CreateKeycloakApiService,
-  type KeycloakApiService,
-} from "src/database/services/keycloak-api-service";
-
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import type { User } from "src/types/keycloak/user";
 /**
- * Response schema for lambda
+ * FIXME: At this moment in KeyCloak its called: severa-user-id (string);
```
Jdallos: Same as in previous file, lets leave a single comment explaining this, perhaps in the User type definition

src/functions/keycloak/list-users/handler.ts
```diff
@@ -27,22 +15,24 @@ const listUsersHandler: APIGatewayProxyHandler = async () => {
 
     const users = await api.getUsers();
```
Jdallos, line 12: TS thinks this is of type User form the getUsers method return type definition. But becuase you are doig "filteredUsers" I suspect its not. See previous comment in other file, but this should be typed correctly so it's clear why we are filtering users.

Also I would change the naming, I think in this case we would usually call it mappedUsers. 

If it seems I've misunderstood something lets discuss further

src/functions/keycloak/list-users/handler.ts
```diff
@@ -27,22 +15,24 @@ const listUsersHandler: APIGatewayProxyHandler = async () => {
...
 
-    const mappedUsers: Response[] = users.map((user) => {
-      const responseUser: Response = {
+    const filteredUsers: User[] = [];
+
+    for (let i = 0; i < users.length; i++) {
```
Jdallos: This should be done with a map, unless there is a good reason not to?

Jdallos (changes requested):

src/functions/keycloak/list-users/handler.ts
dalju005-xamk: Sure, this was note from Ilmo to avoid mapping. There was something with SonarCloud giving negative feedback about mapping.

src/database/services/keycloak-api-service.ts
```diff
@@ -1,12 +1,37 @@
-import type { User } from "../../types/keycloak/user";
 import fetch from "node-fetch";
 
+/**
+ * Interface for a KeycloakProfile. 
+ * 
+ * FIXME: This is from node_modules/keycloak-js/lib/keycloak.d.ts
+ *  This a temporary solution to avoid the error: "Cannot find module 'keycloak-js'"  
+ */
+
```
Jdallos:
```suggestion
```

src/database/services/keycloak-api-service.ts
```diff
@@ -1,12 +1,37 @@
...
+export interface KeycloakProfile {
+	id?: string;
```
Jdallos: Fix indentation

src/database/services/keycloak-api-service.ts
```diff
@@ -1,12 +1,37 @@
...
+	enabled?: boolean;
+	emailVerified?: boolean;
+	totp?: boolean;
+	createdTimestamp?: number;
+	attributes?: Record<string, unknown>;
+}
+/**
+ * Custom Interface for a user in keycloak functions with severaGuid added.
+ */
+
```
Jdallos: There is no need for a line break between the Docs and the thing that is being documented. I keep seeing this

Jdallos:
```suggestion
```

src/database/services/keycloak-api-service.ts
```diff
@@ -36,7 +61,11 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
         );
       }
 
-      return response.json();
+      const users: KeycloakProfile[] = await response.json();
+      return users.map((user) => ({
+        ...user,
+        severaGuid: (user as CustomKeycloakProfile).severaGuid ?? "Users SeveraGuid not found in Keycloak",
```
Jdallos: We don't want to set the severaGuid as a error message string. Let's set it as undefined or null- you decide.
Also Stepan suggested last week we change all references to Guid to Id and I tend to agree. Lets do this here unless you disagree?

src/database/services/keycloak-api-service.ts
```diff
@@ -60,7 +89,11 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
         throw new Error(`Failed to find user with id: ${id}`);
       }
 
-      return response.json();
+      const user: KeycloakProfile = await response.json();
+      return {
+        ...user,
+        severaGuid: (user as CustomKeycloakProfile).severaGuid ?? "Users SeveraGuid not found in Keycloak",
```
Jdallos: Undefined or null same as above, also rename

src/functions/keycloak/find-user/handler.ts
```diff
@@ -24,27 +20,17 @@ const findUserHandler: APIGatewayProxyHandler = async (
       throw new Error("Missing or invalid path parameter: id");
     }
 
-    const userById: User = await api.findUser(queryStringParameters.id);
+    const userById = await api.findUser(queryStringParameters.id);
```
Jdallos, line 22: There is an unnecessary line break between Docs and findUserHandler

src/functions/keycloak/list-users/handler.ts
```diff
@@ -1,10 +1,7 @@
 import { middyfy } from "@libs/lambda";
 import type { APIGatewayProxyHandler } from "aws-lambda";
 import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
-import type { User } from "src/types/keycloak/user";
-/**
- * FIXME: At this moment in KeyCloak its called: severa-user-id (string);
- */
+
```
Jdallos, line 4: No need for a double line break here

src/types/keycloak/user.ts
```diff
@@ -9,4 +9,7 @@ export interface User {
     isActive: boolean;
     severaGuid: string;
     forecastId: number;
-}
\ No newline at end of file
+}
+/**
+ * FIXME: At this moment in KeyCloak its called: severa-user-id (string), in here its severaGuid (string);
```
Jdallos: Lets rename to severaId and remove this comment

Jdallos (changes requested):

src/database/services/keycloak-api-service.ts
dalju005-xamk: Stepan success that we use severaUserGuid. I think that would be fine.

src/database/services/keycloak-api-service.ts
```diff
@@ -6,25 +6,23 @@ import fetch from "node-fetch";
...
+  attributes?: Record<string, unknown>;
 }
 /**
- * Custom Interface for a user in keycloak functions with severaGuid added.
+ * Custom Interface for a user in keycloak functions with severaUserGuid added.
  */
-
 export interface CustomKeycloakProfile extends KeycloakProfile {
-  severaGuid: string;
+  severaUserGuid: string;
```
Jdallos: I said in a previous comment, lets rename all Guid references to Id

src/types/keycloak/user.ts
```diff
@@ -7,9 +7,6 @@ export interface User {
     lastName: string;
     email: string;
     isActive: boolean;
-    severaGuid: string;
+    severaUserGuid: string;
```
Jdallos: Rename

Jdallos (changes requested):

Jdallos (approved):

PR #95, opened 2024-10-22
Feature 91 vacation lambda design

### LINKED ISSUE #91: Vacation Lambda Design
Adding lambda functions according to the vacation request spec design

naStipe:
Lambda functions done with the same structure and functionality as before (as there is no guidance on whether or not it should be working differently now). The lambda functions were checked locally with mock.json file (they were left on git)


src/functions/vacation-request/create-vacation-request/handler.ts
AfoninOleksandr, line 1: Keep the

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (!event.body) {
+        return {
+            statusCode: 400,
+            body: JSON.stringify({ error: "Request body is required." })
+        };
+    }
+
+    const { personId, draft, startDate, endDate, days, type, message, createdBy, createdAt, updatedAt} = event.body;
+
+    if (!personId || !draft || !startDate || !endDate || !days || !type || !message || !createdBy || !createdAt || !updatedAt) {
```
AfoninOleksandr: Could be better to separate the user input data checks so that the user knows what exactly is missing from the request. (not an error just keep in mind)

AfoninOleksandr (approved):
Looks good

.vscode/launch.json
Jdallos, line 1: Do we need this file? If so please explain otherwise remove.

serverless.ts
```diff
@@ -33,6 +33,12 @@ import findQuestionnaireHandler from "@/functions/questionnaire/find-questionnai
 import deleteQuestionnaireHandler from "src/functions/questionnaire/delete-questionnaire";
 import listQuestionnaireHandler from "src/functions/questionnaire/list-questionnaire";
 import updateQuestionnaireHandler from "src/functions/questionnaire/update-questionnaire";
+import createVacationRequestHandler from "src/functions/vacation-request/create-vacation-request";
+import deleteVacationRequestHandler from "src/functions/vacation-request/delete-vacation-request";
+import findVacationRequestHandler from "src/functions/vacation-request/find-vacation-request";
+import listVacationRequestHandler from "src/functions/vacation-request/list-vacation-request";
+import updateVacationRequestHandler from "src/functions/vacation-request/update-vacation-request";
+
```
```suggestion
```
No double line breaks

serverless.ts
Jdallos, line 1: I think this is missing some changes. Look at how a questionnaire DB table is configured here, the same needs to be done for the vacations table and the vacations status table.

src/database/models/vacaionRequestStatus.ts
```diff
@@ -0,0 +1,19 @@
+
+interface VacationRequestStatusModel {
```
Jdallos: Add missing TSDocs- see models/questionnaire.ts for examples

src/database/models/vacaionRequestStatus.ts
```diff
@@ -0,0 +1,19 @@
+
```
Jdallos:
```suggestion
```

src/database/models/vacaionRequestStatus.ts
```diff
@@ -0,0 +1,19 @@
...
+    vacationRequestId: string;
+    status: VacationRequestStatuses;
+    message: string;
+    createdAt: string;
+    updatedAt: string;
+    createdBy: string;
+    updatedBy: string;
+}
+
+enum VacationRequestStatuses {
```
Jdallos: Missing Docs

Jdallos: Is this not being exported?

src/database/models/vacationRequest.ts
```diff
@@ -0,0 +1,32 @@
...
+    createdAt: string;
+    updatedAt: string;
+}
+
+/**
+ * DynamoDB model for vacation type
+ */
+
+enum VacationType {
+    VACATION = "VACATION",
```
Jdallos: Please check your VSCode indentation- all files should be indented using 2 spaces, this file is set to 4 spaces

src/database/models/vacationRequest.ts
```diff
@@ -0,0 +1,32 @@
...
+    message: string;
+    createdBy: string;
+    createdAt: string;
+    updatedAt: string;
+}
+
+/**
+ * DynamoDB model for vacation type
+ */
+
```
Jdallos:
```suggestion
```
Unnecessary line break

src/database/services/index.ts
```diff
@@ -1,4 +1,7 @@
 import createDynamoDBClient from "../client";
 import QuestionnaireService from "./questionnaire-api-service";
+import VacationRequestService from "@database/services/vacation-request-api-service";
 
-export const questionnaireService = new QuestionnaireService(createDynamoDBClient());
\ No newline at end of file
+
```
Jdallos:
```suggestion
```

src/database/models/vacaionRequestStatus.ts
Jdallos, line 1: This seems to be the only reference to vacationRequestStatuses. It's reasonable to just deal with vacation requests for now and add the statuses as a separate feature (they will need their own service). But perhaps this file should be removed for now and added later when we actually add the status functionality?

src/database/services/vacation-request-api-service.ts
```diff
@@ -0,0 +1,100 @@
+import type { DocumentClient } from "aws-sdk/clients/dynamodb";
+
```
Jdallos, line 3:
```suggestion
```

src/database/services/vacation-request-api-service.ts
```diff
@@ -0,0 +1,100 @@
...
+
+const TABLE_NAME = 'VacationRequests';
```
Jdallos: Always use double quotes unless there is actually a reason not to.

src/database/services/vacation-request-api-service.ts
```diff
@@ -0,0 +1,100 @@
...
+
+class VacationRequestService {
```
Jdallos, line 9: MIssing Docs

src/database/services/vacation-request-api-service.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Constructor
+     * @param docClient DynamoDB client
+     */
+    constructor(private readonly docClient: DocumentClient) {}
```
Jdallos: Check indentation

src/database/services/vacation-request-api-service.ts
```diff
@@ -0,0 +1,100 @@
...
+
+
+    /**
+     * Creates a vacation request
+     *
+     * @param vacationRequest vacation request
+     * @returns created vacationRequest
+     */
+    public createVacationRequest = async (vacationRequest: VacationRequestModel): Promise<VacationRequestModel> => {
```
Jdallos: Looks like VacationRequestModel needs to be imported. Keep an eye out for errors/ problems in VSCode which will highlight issues like this

src/functions/vacation-request/create-vacation-request/handler.ts
Jdallos: I think in this case it is ok to return some kind of generic 400 "invalid body" error.

Jdallos: I suggest in the spec review that draft does not need to be required.

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+import type { ValidatedEventAPIGatewayProxyEvent } from "src/libs/api-gateway";
+import type vacationRequestSchema from "src/schema/vacationRequest";
+
+/**
+ * Handler for creating a new vacation request entry in DynamoDB.
+ *
+ * @param event - API Gateway event containing the request body.
+ * @returns Response object with status code
+ */
+export const createVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<typeof vacationRequestSchema> = async (event) => {
```
Jdallos, line 13: Again the indentation is set incorrectly. I won't mention it again but please check all files and update so they have an indentation of 2 spaces

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+    const { personId, draft, startDate, endDate, days, type, message, createdBy, createdAt, updatedAt} = event.body;
+
+    if (!personId || !draft || !startDate || !endDate || !days || !type || !message || !createdBy || !createdAt || !updatedAt) {
+        return {
+            statusCode: 400,
+            body: JSON.stringify({ error: "Some required data is missing !" })
+        };
+    }
+
+    const newVacationRequestId: string = uuidv4();
```
Jdallos: Don't need explict typing here

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+    let vacationRequestResponse: VacationRequestModel | undefined = undefined;
```
Jdallos: I don't see any purpose for doing this (although I see it was done this way for Questionnares).
I would suggest removing this and returning the createdVacationRequest directly (in the body and stringified).

Jdallos: As with all these comments, Let me know if you disagree

src/functions/vacation-request/create-vacation-request/mock.json
Jdallos, line 1: It's reasonable to have a single mock data but it should be in a general folder as this mock data can be helpful when invoking many functions, I'd suggest moving this to tests/serverless/mock.json.

And then we should not commit anymore mock files, rather they should be local only.

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,57 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler} from "aws-lambda";
+import { vacationRequestService } from "src/database/services";
+import { middyfy } from "src/libs/lambda";
+
+/**
+ * Lambda for deleting a vacation request entry from DynamoDB.
+ * 
+ * @param event event
+ */
+
```
Jdallos:
```suggestion
```

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,57 @@
...
+const deleteVacationRequestHandler: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {
+    /*APIGatewayHandler (aka Handler<APIEvent, APIResult>) will take API event and result as params
+    Received API Event, Return API Result */
+
```
Jdallos:
```suggestion
```

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,57 @@
...
+    try {
+        if (!id) {
+            return {
+                statusCode: 400,
+                body: JSON.stringify({
+                    error: "Missing or invalid path parameter: id",
+                }),
+            };
+        }
```
Jdallos: This doesn't need to be in the try.

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        const findVacationRequestById = await vacationRequestService.findVacationRequest(id);
```
Jdallos: I would rename to foundVacationRequestById

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,57 @@
...
+                details: error.message,
+            }),
+        };
+    }
+}
+
+export const main = middyfy(deleteVacationRequestHandler);
+
+/* middy helps to integrate multiple middlewares
+in lambda withoud cluttering the handler logic */
```
Jdallos:
```suggestion
```

src/functions/vacation-request/delete-vacation-request/mock.json
Jdallos, line 1: Lets only have a single mock data file which can be referenced by all functions and changed locally as required. We don't want to clutter the code with mock data but I think it's reasonable in this case to have a single "template" json file as it is needed for invoking locally.

So this file should be removed

src/functions/vacation-request/find-vacation-request/handler.ts
```diff
@@ -0,0 +1,51 @@
...
+import { vacationRequestService } from "src/database/services";
+
+/**
+ * Lambda for finding a vacation request entry from DynamoDB.
+ * 
+ * @param event event
+ * @returns vacation request information as object 
+ * 
+ */
+
```
Jdallos:
```suggestion
```

src/functions/vacation-request/find-vacation-request/handler.ts
```diff
@@ -0,0 +1,51 @@
...
+    try {
+        if (!id){
+            return {
+                statusCode: 400,
+                body: JSON.stringify({
+                    error: "Missing or invalid path parameter: id",
+                })
+            }
+        }
```
Jdallos: Does this need to be in the try?

src/functions/vacation-request/find-vacation-request/handler.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        return {
+            statusCode: 200,
+            body: JSON.stringify(vacationRequestById)
+        }
+    } catch (error) {
+        return {
+            statusCode: 500,
+            body:JSON.stringify({
+                error: "Failed to retrieve vacation requests",
```
Jdallos:
```suggestion
                error: "Failed to retrieve vacation request",
```

src/functions/vacation-request/find-vacation-request/mock.json
Jdallos, line 1: As said in other comments this should be deleted

src/functions/vacation-request/list-vacation-request/mock.json
Jdallos, line 1: Remove this file

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -0,0 +1,71 @@
...
+    draft,
+    startDate,
+    endDate,
+    days,
+    type,
+    message,
+    createdBy,
+    createdAt,
+    updatedAt
+  } = body;
```
Jdallos: Do we need some kind of validation check for these e.g. similar to the create function where is a required attribute is missing we return a 400 "invalid body" response?

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -0,0 +1,71 @@
...
+    return {
+      statusCode: 404,
+      body: `Vacation request ${id} not found`
+    };
+  }
+
+  const vacationRequestUpdates: VacationRequestModel = {
+    id: existingVacationRequest.id,
+    personId: personId,
+    draft: draft,
```
Jdallos: If draft is not required, and is not included in the body, we can assume it's false.

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -0,0 +1,71 @@
...
+    startDate: startDate,
+    endDate: endDate,
+    days: days,
+    type: type,
+    message: message,
+    createdBy: createdBy,
+    createdAt: createdAt,
```
Jdallos: If we are updating perhaps we should not be changing the createdBy or CreatedAt. Does this mean we need an updated by?

src/functions/vacation-request/update-vacation-request/mock.json
Jdallos, line 1: This can be removed

src/schema/vacationRequest.ts
```diff
@@ -0,0 +1,19 @@
+import { Type } from "@sinclair/typebox";
+
+/**
+ * Schema for the Questionnaire table
+ */
+
```
Jdallos, line 13:
```suggestion
```

Jdallos (changes requested):

serverless.ts
```diff
@@ -188,6 +198,19 @@ const serverlessConfiguration: AWS = {
...
+          TableName: "VacationRequests",
+          AttributeDefinitions: [{ AttributeName: "id", AttributeType: "S" }],
+          KeySchema: [{ AttributeName: "id", KeyType: "HASH" }],
+          ProvisionedThroughput: {
+            ReadCapacityUnits: 1,
+            WriteCapacityUnits: 1
+          },
+        }
+      },
       Software: {
```
dalju005-xamk: Remember to ad IAM role for you new table. Like with Software Registery or Questionnaires

dalju005-xamk (approved):

serverless.ts
```diff
@@ -135,6 +140,19 @@ const serverlessConfiguration: AWS = {
...
+            Action: [
+              "dynamodb:DescribeTable",
+              "dynamodb:Query",
+              "dynamodb:Scan",
+              "dynamodb:GetItem",
+              "dynamodb:PutItem",
+              "dynamodb:UpdateItem",
+              "dynamodb:DeleteItem",
+            ],
+            Resource: isLocal ? "*" : "arn:aws:dynamodb:${self:provider.region}:*:table/VacationRequests"
```
Jdallos: I think we should try to define there permissions another way, it seems to be done in a different way in this repo,
https://github.com/Metatavu/elastic-app-functions/blob/develop/serverless.ts

I'll also look into this a bit further.

src/database/models/vacaionRequestStatus.ts
Jdallos: ?

src/database/models/vacaionRequestStatus.ts
Jdallos: ?

src/database/models/vacationRequest.ts
Jdallos: ?

src/functions/vacation-request/create-vacation-request/handler.ts
naStipe: Why is that important here?

src/database/models/vacaionRequestStatus.ts
Jdallos, line 1: Commented files are generally not a great idea, especially in a project that is already bloated with depricated files like this one. I'd suggest removing this, it means if someone else comes to start implementing the vacation statuses functinoality there will hopefully be less confusion.

src/database/models/vacationRequest.ts
```diff
@@ -0,0 +1,38 @@
...
```
Jdallos, line 3:
```suggestion
```

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -0,0 +1,62 @@
...
+        return {
+            statusCode: 201,
+            body: JSON.stringify(createdVacationRequest)
+        };
+    } catch (error) {
+        return {
+            statusCode: 500,
+            body: `Failed to create vacation request entry ${error}`
+        }
+    }
```
Jdallos, line 56: Please check all files have indention set to spaces 2 (and update), this is set to spaces 4

src/functions/vacation-request/create-vacation-request/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
```
Jdallos, line 3: Indentation

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,51 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler} from "aws-lambda";
+import { vacationRequestService } from "src/database/services";
+import { middyfy } from "src/libs/lambda";
+
```
Jdallos, line 4: Fix indentation

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Lambda for deleting a vacation request entry from DynamoDB.
+ * 
+ * @param event event
+ */
+const deleteVacationRequestHandler: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {
+
```
Jdallos, line 12:
```suggestion
```

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,51 @@
...
+        };
+    }
+
+    try {
+        const foundVacationRequestById = await vacationRequestService.findVacationRequest(id);
+        if (!foundVacationRequestById) {
+            return {
+                statusCode: 404,
+                body: JSON.stringify({
+                    error: "Request ${id} not found.",
```
Jdallos:
```suggestion
                    error: "Vacation request with id:  ${id} not found.",
```

src/functions/vacation-request/delete-vacation-request/handler.ts
```diff
@@ -0,0 +1,51 @@
...
+
+        return {
+            statusCode: 204,
+            body: JSON.stringify(""),
+        };
+    } catch (error) {
+        return {
+            statusCode: 500,
+            body: JSON.stringify({
+                error: "Failed to delete request.",
```
Jdallos:
```suggestion
                error: "Failed to delete vacation request.",
```

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Handler for creating a new vacation request entry in DynamoDB.
+ *
+ * @param event - API Gateway event containing the request body.
+ * @returns Response object with status code
+ */
+export const createVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<typeof vacationRequestSchema> = async (event) => {
+    if (!event.body) {
+        return {
+            statusCode: 400,
+            body: JSON.stringify({ error: "Request body is required." })
```
Jdallos: I think all error messages should be updated to refer to Vacation request, rather than Request- I think it's a bit clearer- especially if we think from a UI perspective there are many "Requests" being made

src/functions/vacation-request/delete-vacation-request/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
```
Jdallos, line 3: Indentation

src/functions/vacation-request/find-vacation-request/handler.ts
```diff
@@ -0,0 +1,50 @@
+import { middyfy } from "src/libs/lambda";
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+import { vacationRequestService } from "src/database/services";
+
+/**
+ * Lambda for finding a vacation request entry from DynamoDB.
+ * 
+ * @param event event
+ * @returns vacation request information as object 
+ *
```
Jdallos:
```suggestion
```

src/functions/vacation-request/find-vacation-request/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+    handler: `${handlerPath(__dirname)}/handler.main`,
```
Jdallos: Fix indentation

src/functions/vacation-request/list-vacation-request/handler.ts
```diff
@@ -0,0 +1,29 @@
...
+import type VacationRequestModel from "src/database/models/vacationRequest";
+import { vacationRequestService } from "src/database/services";
+import { middyfy } from "src/libs/lambda";
+
+/**
+ * Labmda for listing all questions from DynamoDB.
+ */
+const listVacationRequestHandler: APIGatewayProxyHandler = async () => {
+  try {
+    const allVacationRequests: VacationRequestModel[] =
```
Jdallos, line 11: Do we need to explicity type here?

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (!existingVacationRequest) {
+    return {
+      statusCode: 404,
+      body: `Vacation request ${id} not found`
+    };
+  }
+
+  const vacationRequestUpdates: VacationRequestModel = {
+    id: existingVacationRequest.id,
+    personId: personId,
```
Jdallos: Would it be reasonable for this personId to also come from the existingVacationRequest, as surely we don't want this to be able to be updated?

Jdallos (changes requested):

serverless.ts
naStipe: I dont see how exactly much different from the way we did it, besides the recources. Could you elaborate on that?

src/functions/vacation-request/create-vacation-request/handler.ts
naStipe: Will do. Although I thought that "request" in this case refers to the API request, rather than to a vacation request. Doesn`t it?

serverless.ts
Jdallos: Here we seem to be specifying the actions array for each resource/ table. The example I linked is setting the same actions for all the resources but without having to repeat the 7 lines of "actions" like we are here. At least I think this is what it is doing, you can check the serverless docs to ensure this what it is doing.

 As we can expect the number of tables to increase it would be nice if we could just add a single line for the resource rather than repeat the actions array for each one. We could also go through this together later

src/functions/vacation-request/create-vacation-request/handler.ts
Jdallos: Ha you are right! Well I guess that's proof there is the possibility for confusion. So this is fine and does not need changing.

Jdallos (changes requested):

src/database/models/vacationRequest.ts
```diff
@@ -3,35 +3,35 @@
...
  * Enumerator for the vacation request statuses
  */
 enum VacationRequestStatuses {
-    PENDING = "PENDING",
-    APPROVED = "APPROVED",
-    REJECTED = "DECLINED"
+  PENDING = "PENDING",
+  APPROVED = "APPROVED",
+  REJECTED = "DECLINED"
 }
```
Jdallos: This is not currently in use. Also when this is needed/ in use we can use the enum which is already provided from the generated files

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -12,51 +12,51 @@ import VacationRequestModel from "@database/models/vacationRequest";
...
-    if (!event.body) {
-        return {
-            statusCode: 400,
-            body: JSON.stringify({ error: "Request body is required." })
-        };
-    }
+  if (!event.body) {
+    return {
+      statusCode: 400,
+      body: JSON.stringify({ error: "Vacation request body is required." })
```
Jdallos: As said in previous comment this is my mistake, please revert

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -29,11 +30,11 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<typeof va
   if (!id) {
     return {
       statusCode: 400,
-      body: "Bad request, missing id"
+      body: "Bad vacation request, missing id"
```
Jdallos: Again my mistake, please revert

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -29,11 +30,11 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<typeof va
...
     };
   }
 
-  if(!personId || !draft || !startDate || !endDate || !days || !type || !message || !createdBy || !createdAt || !updatedAt || !updatedBy) {
+  if(!personId || !draft || !startDate || !endDate || !days || !type || !status || !message || !createdBy || !createdAt || !updatedAt || !updatedBy) {
```
Jdallos: Does this mean you are opting to have status as an attribute within vacation rather than a separate entity?

If so I think this can work but we need to discuss the specific requirements for updating the status before we go ahead with this

src/schema/vacationRequest.ts
```diff
@@ -5,15 +5,15 @@ import { Type } from "@sinclair/typebox";
  */
 
 const vacationRequestSchema = Type.Object({
```
Jdallos, line 17: If you decide to add status to the VacationRequest, wouldn't it also need to be here?

Jdallos (changes requested):

src/database/models/vacationRequest.ts
```diff
@@ -19,12 +20,13 @@ interface VacationRequestModel {
...
 
 /**
- * Enumerator for the vacation request statuses
+ * DynamoDB model for vacation request status
  */
-enum VacationRequestStatuses {
-  PENDING = "PENDING",
-  APPROVED = "APPROVED",
-  REJECTED = "DECLINED"
+interface VacationRequestStatusesModel {
```
Jdallos: Do we need this model though, shouldn't this be described in the spec and then we can just reference the spec. Sometimes the spec and a DB model differ but in this case I don't think there is any reason for them so we can just have a single source defining the status type.

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -24,7 +23,7 @@ export const createVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<ty
   if (!personId || !startDate || !endDate || !days || !type || !message || !createdBy || !createdAt || !updatedAt || !updatedBy) {
     return {
       statusCode: 400,
-      body: JSON.stringify({ error: "Some required data is missing !" })
+      body: JSON.stringify({ error: "Invalid request body! Some data is missing!" })
```
Jdallos:
```suggestion
      body: JSON.stringify({ error: "Invalid request body. Some data is missing." })
```

src/schema/vacationRequest.ts
```diff
@@ -1,15 +1,26 @@
 import { Type } from "@sinclair/typebox";
+import {VacationRequestStatuses} from "@generated/client/model/vacationRequestStatuses";
 
 /**
- * Schema for the Questionnaire table
+ * Schema for each individual status entry in a vacation request
  */
+const vacationRequestStatusSchema = Type.Object({
```
Jdallos, line 7: So the spec will need updating again, we need to remove the vacationRequestStatuses endpoint and update the schema.
Also remember if any of these attributes are optional it should be described as such here.

And related to this, it looks like at least some VacationRequest attributes are described as optional in the spec and I don think some of them should be optional e.g. startDate, createdBy. I should have spotted it previously but might as well get it updated now.

src/schema/vacationRequest.ts
```diff
@@ -1,15 +1,26 @@
...
+  status: Type.Enum(VacationRequestStatuses),
+  message: Type.String(),
+  createdBy: Type.String(),
+  updatedAt: Type.String()
+})
 
+/**
+ * Schema for the vacation request table
+ */
 const vacationRequestSchema = Type.Object({
```
Jdallos, line 17: This should match the spec e.g. draft is optional (or maybe it doesn't need to be optional), but I think this is missing at least personId? And status needs to be added to the spec. 

And just to be clear we don't want id here, and it's not here, so that fine!

Jdallos (approved):

PR #109, opened 2024-11-12
Change eventDate handling

### LINKED ISSUE #108: Remove unused eventDate from Severa/Flextime

dalju005-xamk:
No it uses yesterdays value instead of queryString parameters

### DIFF
--- a/src/database/services/severa-api-service.ts
+++ b/src/database/services/severa-api-service.ts
@@ -1,5 +1,6 @@
 import fetch from "node-fetch";
 import type Flextime from "../models/severa";
+import { DateTime } from "luxon";
 
 /**
  * Interface for a SeveraApiService.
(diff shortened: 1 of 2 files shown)


src/database/services/severa-api-service.ts
```diff
@@ -18,8 +18,13 @@ export const CreateSeveraApiService = (): SeveraApiService => {
     /**
      * Gets flextime by userGUID and eventDate
      */
-    getFlextimeBySeveraUserId: async (severaUserId: string, eventDate: string) => {
-      const url: string = `${baseUrl}/v1/users/${severaUserId}/flextime?eventdate=${eventDate}`;
+    getFlextimeBySeveraUserId: async (severaUserId: string) => {
+
+      const eventDate = new Date();
+      eventDate.setDate(eventDate.getDate() - 1);
+      const formattedEventDate = eventDate.toISOString().split('T')[0];
```
Jdallos: Dates can often have unexpected behaviours, especially when we consider projects potentially being deployed in different time zones. So In the majority of cases anything requiring formatting or editing of a date we should use a date time library.

Currently we are using luxon, it's already installed in this project and has good documentation. So do the above: setting the date and formatting, using luxon

src/database/services/severa-api-service.ts
```diff
@@ -18,8 +18,13 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+
+      const url: string = `${baseUrl}/v1/users/${severaUserId}/flextime?eventdate=${formattedEventDate}`;
```
Jdallos: No need to explicit typing here

Jdallos (changes requested):

PR #114, opened 2024-11-18
Feature 92 severa lambda design for sprint view

Hanhz2707:


src/functions/forecast/list-allocations/mockData6.json
```diff
@@ -0,0 +1,6 @@
+{
+    "headers": {
+      "Content-Type": "application/json"
+    },
```
dalju005-xamk: I think you should not try to merge mockData to develop

src/database/models/flextime.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * What Flextime returns from Severa.
+ */
+interface Flextime {
+  totalFlextimeBalance: number | null;
```
dalju005-xamk: Why is this added here ? It should already be in develop ?

src/functions/forecast/list-allocations/mockData6.json
Hanhz2707: done

src/database/models/flextime.ts
Hanhz2707: hmmm... i actually have no idea

src/database/services/severa-api-service.ts
```diff
@@ -1,20 +1,26 @@
...
 import { DateTime } from "luxon";
 
 /**
  * Interface for a SeveraApiService.
  */
 export interface SeveraApiService {
   getFlextimeBySeveraUserId: (severaUserId: string) => Promise<Flextime>;
+  getResourceAllocation: (severaUserId: string) => Promise<ResourceAllocationModel[]>;
+  getPhasesBySeveraProjectId: (severaProjectId: string) => Promise<Phase[]>;
+  getWorkHoursBySeveraId: ( severaProjectId?: string, severaUserId?: string, severaPhaseId?: string, startDate?: string, endDate?: string) => Promise<WorkHours[]>;
```
naStipe: ResourceAllocationModel and Phase. Is it possible to either have "Model" in all of them or in none to keep it constant in the project?

naStipe (approved):

dalju005-xamk (approved):

src/database/models/phase.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Severa model for project
+ */
+interface PhaseProjectSubModel{
+  severaProjectId: string;
+  name: string;
+  isClosed: boolean;
+}
+
+export default PhaseModel;
```
Jdallos: I would suggest moving these types into the types folder, perhaps creating a types/severa/pase folder. As we are not doing any DB operations here and so they don't need to be in the database/models folder.

Similar should be done for models/resourceAllocartion and models/workhours

src/database/models/phase.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Severa model for Phases
+ */
+interface PhaseModel {
+  severaPhaseId: string;
+  name: string;
+  isCompleted: boolean;
+  workHoursEstimate: number;
+  startDate: Date;
+  deadLine: Date;
```
Jdallos: Do you think endDate might be a clearer name than deadLine? Assuming that's what deadline is. It would be more in line with other schemas in the project.

src/database/models/phase.ts
```diff
@@ -0,0 +1,23 @@
...
+  project: PhaseProjectSubModel;
+}
+
+/**
+ * Severa model for project
```
Jdallos: Can you clarify this Doc (just a couple of words more) a little as its not just a project is it, its a phase project sub module- what relevance does that have?

src/database/models/resourceAllocation.ts
```diff
@@ -0,0 +1,38 @@
+/*
```
Jdallos: Fix indentation in this file, should be spaces 2

src/database/models/resourceAllocation.ts
```diff
@@ -0,0 +1,38 @@
...
+ *  ResourceAllocation model
+*/
+interface ResourceAllocationModel {
+    severaProjectId: string;
+    allocationHours: number;
+    calculatedAllocationHours: number;
+    phase: PhaseSubModel;
+    users: UserSubModel;
+    projects: ResourceAllocationProjectSubModel;
```
Jdallos: Does an allocation always have a phase and a project?
These considerations might become clearer when you start working with the UI. If you think this model is correct then it's fine by me

Jdallos: Looking at the severa Docs it looks like the attributes of the ResourceAllocationPhaseSubmodel are all nullable. Which makes me think the answer to my above question is that an allocation might not always have a phase.

However If in this example the phaseSubmodel from the severa documentation (assuming it's the same as ResourceAllocationPhaseSubmodel)  can have its attributes as nullable (optional) shouldn't that be reflected in your interface below PhaseSubModule? I'd recommend double checking all your severa interfaces to make sure they reflect when attributes are optional. This is likely to also require changes to the spec. Please correct me if I'm misunderstanding something


 Related it might be clearer if you were consistent in the naming, e.g. why have you called it resourceAllocationProjectSubModel, but just phase/userSubModel. We should be consistant in naming.

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +45,86 @@ export const CreateSeveraApiService = (): SeveraApiService => {
           `Failed to fetch flextime: ${response.status} - ${response.statusText}`,
         );
       }
+      return response.json();
+    },
+
+    /** 
+     * Get resource allocation by userId
```
Jdallos, line 54: Missing @param docs
Also the function is not indented in line with the TSDocs as they should be, please fix

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +45,86 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        if (!response.ok) {
+            throw new Error(
+                `Failed to fetch resource allocation: ${response.status} - ${response.statusText}`,
+            );
+        }
+        return response.json();
+    },
+
+    /**
+     * Gets phases by userId
```
Jdallos: Missing @param doc

Jdallos: Description is wrong, isn't it by project Id

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +45,86 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+      if (!response.ok) {
+        throw new Error(
+          `Failed to fetch phases: ${response.status} - ${response.statusText}`,
+        );
+      }
+      return response.json();
+    },
+
+    /**
+     * Gets work hours by userId
```
Jdallos: Missing @params Docs

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +45,86 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+    getWorkHoursBySeveraId: async (severaProjectId?:string, severaUserId?: string, severaPhaseId?: string, startDate?: string, endDate?: string) => {
+      let url: string;
 
+      if (severaProjectId) {
+        url = `${baseUrl}/v1/projects/${severaProjectId}/workhours`;
+      } else if (severaUserId) {
+        url = `${baseUrl}/v1/users/${severaUserId}/workhours`;
+      } else {
+        url = `${baseUrl}/v1/workhours`;
+      }
```
Jdallos: What about if user passes multiple params e.g. project Id and user Id?
I suspect this needs expanding, perhaps using a switch might help to keep this clean.

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +45,86 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+      
+      const queryParams = new URLSearchParams();
+      if (startDate) queryParams.append("startDate", startDate);
+      if (endDate) queryParams.append("endDate", endDate);
+      if (queryParams.toString()) {
+        url += `?${queryParams.toString()}`;
+      }
```
Jdallos: Maybe the switch and this logic should be extracted into a seperate function e.g. constructWorkoutUrl or something

src/database/services/severa-api-service.ts
Jdallos: Maybe the description should be just get workhours as its not necessarily by userId

src/database/services/severa-api-service.ts
```diff
@@ -51,6 +136,9 @@ export const CreateSeveraApiService = (): SeveraApiService => {
  * @returns Access token as string
  */
 const getSeveraAccessToken = async (): Promise<string> => {
+  if (process.env.IS_OFFLINE) {
+    return "test-token";
+  }
```
Jdallos: I think this was removed in a recent PR, does it need to be added back in?

src/database/services/severa-api-service.ts
```diff
@@ -59,7 +147,7 @@ const getSeveraAccessToken = async (): Promise<string> => {
   const requestBody = {
     client_id: client_Id,
     client_secret: client_Secret,
-    scope: "users:read",
+    scope: "projects:read, resourceallocations:read, hours:read",
```
Jdallos: Seems to me you have removed the users: read from scope. Are you sure this won't affect other functions?

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +45,86 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        url = `${baseUrl}/v1/users/${severaUserId}/workhours`;
+      } else {
+        url = `${baseUrl}/v1/workhours`;
+      }
+      
+      const queryParams = new URLSearchParams();
+      if (startDate) queryParams.append("startDate", startDate);
+      if (endDate) queryParams.append("endDate", endDate);
+      if (queryParams.toString()) {
+        url += `?${queryParams.toString()}`;
```
Jdallos: You could look into using the URL class to construct the Url, might be better then url +=

src/functions/severa/list-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+import { middyfy } from "src/libs/lambda";
+import type WorkHours from "@database/models/workHours";
+import { FilterUtilities } from "src/libs/filter-utils";
+
+/**
+ * Handler for getting work hours from Severa REST API.
+ *
+ * @param event - API Gateway event containing the userId, projectId, and phaseId.
+ */
+export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
```
Jdallos: File is named list but the handler is get. I would suggest renaming the function to listWorkHoursHandler and the file to lsit-workhours. This will require making the relevant changes to the serverless.ts

src/functions/severa/list-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+  const { severaUserId, severaProjectId, severaPhaseId } = event.pathParameters;
+  const { startDate, endDate } = event.queryStringParameters || {};
+
+  try {
+    const api = CreateSeveraApiService();
+    const response: WorkHours[] = await api.getWorkHoursBySeveraId(severaProjectId, severaUserId, severaPhaseId, startDate, endDate);
+
+    const filteredWorkHours = response.filter((workHours:any) => {
```
Jdallos: Don't we have an actual type for workHours here, rather than just any?

src/functions/severa/list-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+      return (
+        (!severaUserId || FilterUtilities.filterByUserSevera(workHours.user?.guid, severaUserId)) &&
+        (!severaPhaseId || FilterUtilities.filterByPhaseSevera(workHours.phase?.guid, severaPhaseId)) &&
+        (!startDate || FilterUtilities.filterByDateSevera(workHours.startDate, startDate)) &&
+        (!endDate || FilterUtilities.filterByDateSevera(workHours.endDate, endDate))
```
Jdallos: I'm confused I thought you were filtering the workhours using the functionality from Severa, calling a particular endpoint based on params provided. I think this filtering needs to be cleaned up some more. Perhaps we should discuss if you need some ideas. But currently it seems a bit strange to pass all the params to the request (to filter)(and then to filter the  response again

Jdallos: Also the filtering itself is a little unclear, I wonder if it could be broken down and simplified rather than been a single large condition.

src/functions/severa/list-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+      );
+    });
+
+    const mappedWorkHours = filteredWorkHours.map((workHours:any) => ({
```
Jdallos: Does work hours have to be types as any here?

src/functions/severa/list-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,63 @@
...
+      },
+      phase: {
+        severaPhaseId: workHours.phase?.guid,
+        name: workHours.phase?.name,
+      },
+      description: workHours.description,
+      eventDate: workHours.eventDate,
+      quantity: workHours.quantity,
+      startTime: workHours.startTime,
+      endTime: workHours.endTime,
```
Jdallos: Looks to me like some of these attributes are not detailed in the spec. Rememeber the spec must describe what the lambda is doing. It's fine to update the spec if these additional attributes are needed

src/functions/severa/list-phases-by-project/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!severaProjectId) {
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ message: "Project Id is required" }),
+      };
+    }
+
+    const response = await api.getPhasesBySeveraProjectId(severaProjectId);
+
+    const phases =  response.map((item: any) => ({
```
Jdallos: Would phase be a better name than item? Don't we have a type for it rather than any?

src/functions/severa/list-phases-by-project/handler.ts
```diff
@@ -0,0 +1,50 @@
+import type { APIGatewayProxyHandler } from "aws-lambda";
+import { middyfy } from "src/libs/lambda";
+import { CreateSeveraApiService } from "src/database/services/severa-api-service";
+
+/**
+ * Handler for getting Phases by project from Severa REST API.
+ *
+ * @param event - API Gateway event containing the userId.
+ */
+export const getPhasesHandler: APIGatewayProxyHandler = async (event) => {
```
Jdallos: Consistent naming again, can be called list-phases-by-project as is but the function/ handler name should be updated and this updated will need to also be reflected in the serverless.ts

src/functions/severa/list-phases-by-project/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "get",
+        path: "/severa/projects/{severaProjectId}/phaseswithhierarchy",
```
Jdallos: couldn't this endpoint just be phases, rather than phaseswithheirarchy? It's also just phases in the spec

src/functions/severa/list-resource-allocations-by-user/handler.ts
```diff
@@ -0,0 +1,55 @@
+import type { APIGatewayProxyHandler } from "aws-lambda";
+import { CreateSeveraApiService } from "src/database/services/severa-api-service";
+import { middyfy } from "src/libs/lambda";
+
+/**
+ * Handler for getting resourceAllocation by user from Severa REST API.
+ * 
+ * @param event - API Gateway event containing the userId.
+ */
+export const getResourceAllocationHandler: APIGatewayProxyHandler = async (event) => {
```
Jdallos: Consistent naming as described for the previous lambdas, rename handler to listResourceAllocationsHandler.

src/functions/severa/list-resource-allocations-by-user/handler.ts
```diff
@@ -0,0 +1,55 @@
...
+        if(!severaUserId) {
+            return {
+                statusCode: 400,
+                body: JSON.stringify({ message: "User Id is required" }),
+            };
+        }
+
+        const response = await api.getResourceAllocation(severaUserId);
+
+        const resourceAllocation = response.map((item:any) => ({
```
Jdallos: resourceAllocation, not item. Also avoid any type where possible

src/functions/severa/list-resource-allocations-by-user/handler.ts
```diff
@@ -0,0 +1,55 @@
...
+            severaResourceAllocationId: item.guid,
```
Jdallos: This is severaProjectId in the spec, so the spec probably needs fixing

src/functions/severa/list-resource-allocations-by-user/handler.ts
```diff
@@ -0,0 +1,55 @@
+import type { APIGatewayProxyHandler } from "aws-lambda";
```
Jdallos: Fix indentation in this file, check all your files and your default VSCode settings, all indentation should be set at spaces 2

src/functions/severa/list-resource-allocations-by-user/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "get",
+        path: "/severa/users/{severaUserId}/resourceallocations/allocations",
```
Jdallos: This does not match the spec, the final endpoint can just be resourceallocations as detailed in the spec.

Jdallos: Also in the spec its severa/userid/resourcealloactions, I think it should be severa/users/userId/resourceallocations. LEave a comment if this needs clarifying

Jdallos: While making this change to the spec I would also suggest changing the severa/{severaUserId}/flextime to be severa/users/{severaUserId}/flextime. It might be worth discussing this change with Aleksi also, I suggest it to make the endpoints clearer, more structured and follow typical REST convention

src/libs/filter-utils.ts
```diff
@@ -32,6 +32,21 @@ export namespace FilterUtilities {
     return true;
   }
 
+
```
Jdallos: Avoid double line breaks

src/libs/filter-utils.ts
```diff
@@ -32,6 +32,21 @@ export namespace FilterUtilities {
...
+  /**
+   * Compares severa dates to target dates
+   * 
+   * @param startDate Start date from Severa
+   * @param endDate End date from Severa
+   * @returns
```
Jdallos: This @returns can be removed

src/libs/filter-utils.ts
```diff
@@ -32,6 +32,21 @@ export namespace FilterUtilities {
...
+   * @param startDate Start date from Severa
+   * @param endDate End date from Severa
+   * @returns 
+   */
+  export const filterByDateSevera = (startDate?: string, endDate?: string): boolean => {
+    if (startDate === null || endDate === null) {
+      return false;
+    }
+    return true;
+  }
```
Jdallos: Is this actually doing any filtering by date? I'm also not sure how this is being used in the getWorkHoursHandler but I also left some comments regarding changes therre.

src/libs/filter-utils.ts
```diff
@@ -32,6 +32,21 @@ export namespace FilterUtilities {
...
 
+
+  /**
+   * Compares severa dates to target dates
+   * 
+   * @param startDate Start date from Severa
+   * @param endDate End date from Severa
+   * @returns 
+   */
+  export const filterByDateSevera = (startDate?: string, endDate?: string): boolean => {
```
Jdallos: Don't think the explicit return type is necessary here.

src/libs/filter-utils.ts
```diff
@@ -43,10 +58,23 @@ export namespace FilterUtilities {
...
+   * @param targetProjectId Project id to compare
+   * @returns 
+   */
+  export const filterByProjectSevera = (severaProjectId?: string, targetProjectId?: string) : boolean => {
+    if(severaProjectId === null || targetProjectId === null) {
+      return false;
+    }
+    return severaProjectId === targetProjectId;
+  }
+
```
Jdallos: This function appears to be unused, if so it should be removed.

src/libs/filter-utils.ts
```diff
@@ -58,10 +86,23 @@ export namespace FilterUtilities {
...
-
     return true;
   }
 
+  /**
+   * Compares severa User id to targetUser id
+   * 
+   * @param severaUserId User id from Severa
+   * @param targetUserId User id to compare
+   * @returns
```
Jdallos: Remove unnecessary @returns doc and the explicit return typing. I'm also not sure that this filter method should be necessary after getWorkHoursHandler restructuring.
Even if it is needed as the logic is quite short I wonder if it wouldn't just be clearer to do it inline.

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +45,86 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+          `Failed to fetch phases: ${response.status} - ${response.statusText}`,
+        );
+      }
+      return response.json();
+    },
+
+    /**
+     * Gets work hours by userId
+     */
+    getWorkHoursBySeveraId: async (severaProjectId?:string, severaUserId?: string, severaPhaseId?: string, startDate?: string, endDate?: string) => {
```
Jdallos: SeveraPhaseId is currently unused in this function

src/libs/filter-utils.ts
```diff
@@ -73,4 +114,18 @@ export namespace FilterUtilities {
...
   }
+
+  /**
+   * Compares severa phase id to targetPhase id
+   * 
+   * @param severaPhaseId Phase id from Severa
+   * @param targetPhaseId Phase id to compare
+   * @returns 
+   */
+  export const filterByPhaseSevera = (severaPhaseId: string, targetPhaseId: string): boolean => {
```
Jdallos: Similar comments here as for the other filter funcitons, fix Docs, return typing and we can figure out if this filtering is necessary if we can filter the data using severaApi params

Jdallos (changes requested):

src/database/services/severa-api-service.ts
naStipe: Is it worth creating several functions and requesting different tokens when we need different scopes, to not give more scopes than needed each time?

src/database/models/phase.ts
Hanhz2707: So as far as my understanding, phases are inside of projects. This model project was for the project which include the phases according with the interface above it.

src/database/services/severa-api-service.ts
Hanhz2707: If user passes 1st params such as projectID, it will use the url workHours for projects. Then, within getWorkHoursHandler, if user passes 2nd params such as userID, getWorkHoursHandler than will fetch the workHours from that user in that project. if user passes 3rd params such as phaseID, getWorkHoursHandler will do the same as before to get that data. However, it's exactly what you said that it is filtering twice which is not good. Another filtering method is in developing and will be discussed with you later on

src/database/models/phase.ts
Hanhz2707: yes it sounds cleared, however, i noticed it was giving me some errors but i couldn't remember for now as well as it is deadLine according to severa Docs. I will try this today and get back to u for a discussion

src/database/services/severa-api-service.ts
```diff
@@ -24,6 +24,9 @@ export const CreateSeveraApiService = (): SeveraApiService => {
   return {
     /**
      * Gets flextime by severaUserId
+     * 
+     * @param severaUserId  Severa user id
+     * @returns Flextime of an user
```
Jdallos:
```suggestion
     * @returns Flextime for a user
```

src/database/services/severa-api-service.ts
```diff
@@ -50,6 +53,9 @@ export const CreateSeveraApiService = (): SeveraApiService => {
 
     /** 
      * Get resource allocation by userId
+     * 
+     * @param severaUserId Severa user id
+     * @returns Resource allocation of an user
```
Jdallos:
```suggestion
     * @returns Resource allocation for a user
```

src/database/services/severa-api-service.ts
```diff
@@ -93,33 +104,20 @@ export const CreateSeveraApiService = (): SeveraApiService => {
 
     /**
      * Gets work hours by userId
+     * 
+     * @param url url for fetching work hours
```
Jdallos:
```suggestion
     * @param url custom url for fetching work hours depending on query parameter requirements
```

src/database/services/severa-api-service.ts
```diff
@@ -93,33 +104,20 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
-      const queryParams = new URLSearchParams();
-      if (startDate) queryParams.append("startDate", startDate);
-      if (endDate) queryParams.append("endDate", endDate);
-      if (queryParams.toString()) {
-        url += `?${queryParams.toString()}`;
-      }
-
-      const response = await fetch(url, {
+    getWorkHoursBySeveraId: async (url: string) => {
+      const response = await fetch(`${baseUrl}/v1/${url}`, {
```
Jdallos: Maybe we should rename url, to something like endpointPath or whatever you think would be better

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+      } else {
+        url = "workhours";
+      }
+    
+      const queryParams = new URLSearchParams();
+    
+      if (startDate) queryParams.append("startDate", startDate);
+      if (endDate) queryParams.append("endDate", endDate);
+    
+      return queryParams.toString() ? `${url}?${queryParams.toString()}` : url;
```
Jdallos: As mentioned in a previous review could you now use somekind of Url class here to build the url, rather than needing to handle it as a string here when combining the queryParams?

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const url = buildWorkHoursUrl(severaProjectId, severaUserId, startDate, endDate);
+    
+    const response = await api.getWorkHoursBySeveraId(url);
+
+    const filteredWorkHours = response.filter((workHours:SeveraResponse) => {
+      if (severaProjectId && severaUserId) {
+        const filteredWorkHoursUserProject = FilterUtilities.filterByUserSevera(workHours.user?.guid, severaUserId);
+        if (!filteredWorkHoursUserProject) return false ;
+      }
```
Jdallos, line 60: Might be worth adding a comment here to make it clear that in the case of there being severaProjectId and severaUserId it will have been filtered by project from the severaApi (due to the building of workhours url).

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+      if (endDate) queryParams.append("endDate", endDate);
+    
+      return queryParams.toString() ? `${url}?${queryParams.toString()}` : url;
+    };
+
+    const url = buildWorkHoursUrl(severaProjectId, severaUserId, startDate, endDate);
+    
+    const response = await api.getWorkHoursBySeveraId(url);
+
+    const filteredWorkHours = response.filter((workHours:SeveraResponse) => {
```
Jdallos:
```suggestion
    const filteredWorkHours = response.filter((workHours: SeveraResponse) => {
```

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+    
+      if (severaPhaseId) {
+        const filteredWorkHoursPhase = FilterUtilities.filterByPhaseSevera(workHours.phase?.guid, severaPhaseId);
+        if (!filteredWorkHoursPhase) return false ;
+      }
+
+      return true;
+    });
+
+    const mappedWorkHours = (severaData : SeveraResponse[]) :WorkHours[] => {
```
Jdallos: Missing Docs

Jdallos:
```suggestion
    const mappedWorkHours = (severaData : SeveraResponse[]): WorkHours[] => {
```

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+      return severaData.map((workHours) => ({
```
Jdallos: This could be changed to an implicit return e.g. remove the return, replace the function curleyBrackets with parenthesis

src/functions/severa/get-phases-by-project/handler.ts
```diff
@@ -22,19 +24,23 @@ export const getPhasesHandler: APIGatewayProxyHandler = async (event) => {
...
     const response = await api.getPhasesBySeveraProjectId(severaProjectId);
 
-    const phases =  response.map((item: any) => ({
-      severaPhaseId: item.guid,
-      name: item.name,
-      isCompleted: item.isCompleted,
-      workHoursEstimate: item.workHoursEstimate,
-      startDate: item.startDate,
-      deadLine: item.deadline,
+    const mappedPhases = (severaData : SeveraResponsePhases[]) :PhaseModel[] => {
```
Jdallos:
```suggestion
    const mappedPhases = (severaData: SeveraResponsePhases[]): PhaseModel[] => {
```

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -0,0 +1,62 @@
...
+
+try {
+    const api = CreateSeveraApiService();
+
+    if(!severaUserId) {
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ message: "User Id is required" }),
+      };
+    }
```
Jdallos: Do validation at the earliest point possible, I'd move this straight after destructuring the severaUserId

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const response = await api.getResourceAllocation(severaUserId);
+
+    const mappedResourceAllocations = (severaData : SeveraResponseResourceAllocation[]) :ResourceAllocationModel[] => {
```
Jdallos:
```suggestion
    const mappedResourceAllocations = (severaData : SeveraResponseResourceAllocation[]): ResourceAllocationModel[] => {
```

Jdallos: TSDOcs- explain why we are mapping

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -0,0 +1,62 @@
...
+      return severaData.map((resourceAllocation) => ({
```
Jdallos: Could be implicit return

src/tests/serverless/mockData.json
Jdallos, line 1: Generally we don't want to commit mockData, but I think for this repo it is useful to have at least one .json available. So this can remain, but lets make sure no more get committed

src/tests/serverless/mockData.json
```diff
@@ -1,7 +1,9 @@
...
-    },
-    "body": "{\"name\": \"Frederic\"}",
-    "queryStringParameters": { "startDate": "2023-11-03", "endDate": "2021-11-03" }
-}
+  "headers": {
+    "Content-Type": "application/json"
+  },
+  "body": "{\"name\": \"Fredasdasderic\"}",
+  "pathParameters": {
+      "severaUserId": "83ebc1fe-16a4-b070-975f-e21c4334197b"
```
Jdallos: Perhaps a more generic name e.g. TestUser. Also is this a real severUserId? If it is it should be removed and perhaps replaced with [insert severa id here].
Although this isn't a big deal, remember to be extra careful with committing any id's that refer to actual data in the future.

src/types/severa/phase/phase.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+  
+  /**
+   * Interface for PhaseProjectModel
+   */
+  interface PhaseProjectModel{
+    severaProjectId: string;
+    name: string;
+    isClosed: boolean;
+  }
```
Jdallos: Fix indentation here, I also think your vsCode spacing is wrong for this file.

src/types/severa/phase/severaResponsePhases.ts
```diff
@@ -0,0 +1,15 @@
+interface SeveraResponsePhases {
```
Jdallos, line 4: TSDOcs, missing

src/types/severa/resourceAllocation/resourceAllocation.ts
```diff
@@ -0,0 +1,38 @@
+/*
```
Jdallos: Check indentation settings

src/types/severa/resourceAllocation/resourceAllocation.ts
```diff
@@ -0,0 +1,38 @@
...
+    allocationHours: number;
+    calculatedAllocationHours: number;
+    phase?: ResourceAllocationPhaseModel;
+    users: ResourceAllocationUserModel;
+    projects?: ResourceAllocationProjectModel;
+}
+
+/*
+ * Interface for ResourceAllocationPhaseModel
+*/
```
Jdallos: Some of these TSDocs seem misaligned

src/types/severa/resourceAllocation/severaResponseResourceAllocation.ts
```diff
@@ -0,0 +1,21 @@
+
+interface SeveraResponseResourceAllocation {
```
Jdallos, line 4: MIssing TSDOcs

src/types/severa/workHour/severaResponseWorkHours.ts
```diff
@@ -0,0 +1,23 @@
+interface SeveraResponseWorkHours {
```
Jdallos, line 4: Missing TSDocs and check indentation

src/types/severa/workHour/workHour.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
Jdallos, line 1: Check indentation settings and fix indentation, these should all be indented the same amount but they are not.

Jdallos (changes requested):

src/database/services/severa-api-service.ts
```diff
@@ -1,23 +1,32 @@
 import fetch from "node-fetch";
 import type Flextime from "../models/severa";
+import * as process from "node:process";
```
Jdallos: I don't think this import is necessary and can be removed. I've created another task in trello for ensuring envs are imported using envalid but as there are multiple other examples of envs being imported in similar way you are this can be a task on it's own. Feel free to have a look at it if you want but no need for the changes to be done in this PR

src/database/services/severa-api-service.ts
Jdallos: I still think this description should just be gets work hours

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +48,81 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        );
+      }
+      return response.json();
+    },
+
+    /**
+     * Gets work hours by userId
+     * 
+     * @param url custom url for fetching work hours depending on queryparameters requirements
+     * @returns Work hours of a user with multiple queryparameters
```
Jdallos: Is it always work hours of a user though?

Jdallos:
```suggestion
     * @returns Work hours
```

src/database/services/severa-api-service.ts
```diff
@@ -39,7 +48,81 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+     */
+    getWorkHoursBySeveraId: async (endpointPath: string) => {
```
Jdallos: Should this method name just be getWorkHours?

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @param event - API Gateway event containing the userId, projectId, and phaseId.
+ */
+export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
+  const { severaUserId, severaProjectId, severaPhaseId } = event.pathParameters || {};
+  const { startDate, endDate } = event.queryStringParameters || {};
+
+  try {
+    const api = CreateSeveraApiService();
+
```
Jdallos, line 19:
```suggestion
```

Jdallos: No double line breaks

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,116 @@
...
+      } else {
+        endpointPath = "workhours";
+      }
+
+      const customUrl = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/${endpointPath}`);
+        
+      if (startDate) customUrl.searchParams.append("startDate", startDate);
+      if (endDate) customUrl.searchParams.append("endDate", endDate);
+    
+      return customUrl.toString();
```
Jdallos: I believe you could actually pass the Url instance to the fetch method.

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,116 @@
...
+    };
+
+    const url = buildWorkHoursUrl(severaProjectId, severaUserId, startDate, endDate);
+
```
Jdallos:
```suggestion
```

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -0,0 +1,116 @@
...
+          name: workHours.phase?.name,
+        },
+        description: workHours.description,
+        eventDate: workHours.eventDate,
+        quantity: workHours.quantity,
+        startTime: workHours.startTime,
+        endTime: workHours.endTime,
+      }))
+    );
+
```
Jdallos, line 98:
```suggestion
```

Jdallos: No double line breaks

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -0,0 +1,69 @@
...
+          name: resourceAllocation.users?.name,
+        },
+        projects: {
+          severaProjectId: resourceAllocation.projects?.guid,
+          name: resourceAllocation.projects?.name,
+          isInternal: resourceAllocation.projects?.isInternal,
+        },
+      }))
+  );
+
```
Jdallos, line 53:
```suggestion
```

Jdallos: No double line breaks

src/types/severa/workHour/severaResponseWorkHours.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Interface for Severa Response Work Hour.
+ */
+interface SeveraResponseWorkHours {
+  guid: string;
+  user: {
+    guid: string;
+    name: string;
+  };
```
Jdallos: No need for the semi-colon here after the closing curly brackets, please remove all

src/types/severa/workHour/workHour.ts
```diff
@@ -0,0 +1,41 @@
...
+interface WorkHoursModel {
+severaWorkHoursId: string;
+user: WorkHoursUserModel;
+project: WorkHoursProjectModel;
+phase: WorkHoursPhaseModel;
+description: string;
+eventDate: string;
+quantity: number;
+startTime?: string;
+endTime?: string;
```
Jdallos: Indentation for all interfaces, the content of the interface should be indented as it is in all the other types files

Jdallos (approved):

Jdallos (approved):

PR #132, opened 2024-12-02
Added optional query parameter

### LINKED ISSUE #130: Adding optional path parameter to vacation request function
I will add an optional query parameter "userId" to the listVacationRequests function to be able to fetch vacation requests only by their "userId"

naStipe:


src/functions/vacation-request/list-vacation-request/handler.ts
```diff
@@ -6,15 +6,27 @@ import { middyfy } from "src/libs/lambda";
 /**
  * Labmda for listing all vacation requests from DynamoDB.
  */
-const listVacationRequestHandler: APIGatewayProxyHandler = async () => {
+const listVacationRequestHandler: APIGatewayProxyHandler = async (event) => {
+  const userId = event.queryStringParameters?.userId;
   try {
-    const allVacationRequests: VacationRequestModel[] =
-      await vacationRequestService.listVacationRequests();
+    if (userId) {
```
dalju005-xamk, line 12: What if !userIf ? Do we need to consider that ?

src/functions/vacation-request/list-vacation-request/handler.ts
```diff
@@ -6,15 +6,27 @@ import { middyfy } from "src/libs/lambda";
...
+      };
+    }
 
-    return {
-      statusCode: 200,
-      body: JSON.stringify(allVacationRequests),
-    };
   } catch (error) {
     return {
       statusCode: 500,
```
dalju005-xamk, line 30: Should there be error message ?

src/functions/vacation-request/list-vacation-request/handler.ts
naStipe: If it`s !userId then we skip this condition and make a request for all vacation requests on the next lines

src/functions/vacation-request/list-vacation-request/handler.ts
naStipe: Why shouldn`t it? It`s always been there so I just didn`t think I need to do anything with it.

McDuuck (approved):

src/database/services/vacation-request-api-service.ts
```diff
@@ -43,35 +44,50 @@ class VacationRequestService {
...
    *
    * @returns list of vacation requests
    */
-  public listVacationRequests = async (): Promise<VacationRequestModel[]> => {
+  public listVacationRequests = async (userId?: string): Promise<VacationRequestModel[]> => {
+    if (userId) {
+      const filteredResult = await this.docClient
+        .scan({
+          TableName: TABLE_NAME,
+          FilterExpression: "user = :userId",
```
Jdallos: Is it possible for the filterExpression to be undefined/ null, that would allow the userId condition to be handled in a single line rather than needing to have filteredResult or result

src/functions/vacation-request/list-vacation-request/handler.ts
```diff
@@ -6,22 +6,32 @@ import { middyfy } from "src/libs/lambda";
 /**
  * Labmda for listing all vacation requests from DynamoDB.
  */
-const listVacationRequestHandler: APIGatewayProxyHandler = async () => {
+const listVacationRequestHandler: APIGatewayProxyHandler = async (event) => {
+  const userId = event.queryStringParameters?.userId;
   try {
+    if (userId) {
```
Jdallos, line 12: If userId can be undefined in the service as mentioned in the previous comment then we wouldn't need the separate requests here. It might be that filterExpression can't be undefined in which case you can't make these changes which is fine.

I think it would be best (if this structure needs to remain) if you declared const vacationRequests: VacationRequestModel[] = []
And then make the conditional request, update the declared vacationRequests and then return in the same return

Jdallos (approved):

PR #141, opened 2024-12-09
Feature 126 severa opt in

### LINKED ISSUE #126: Severa opt in

rimz178:


src/database/services/keycloak-api-service.ts
```diff
@@ -13,14 +13,15 @@ export interface CustomKeycloakProfile extends KeycloakProfile {
 export interface KeycloakApiService {
   getUsers: () => Promise<CustomKeycloakProfile[]>;
   findUser: (id: string) => Promise<CustomKeycloakProfile>;
+  updateUserAttribute: (id: string, attribute: Record<string, string[]> ) => Promise<void>;
```
Hanhz2707:
```suggestion
  updateUserAttribute: (id: string, attribute: Record<string, string[]>) => Promise<void>;
```

src/functions/keycloak/update-user/handler.ts
```diff
@@ -0,0 +1,46 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import { middyfy } from "src/libs/lambda";
+
+/**
+ *
```
Hanhz2707: Missing description here

src/database/services/keycloak-api-service.ts
```diff
@@ -13,14 +13,15 @@ export interface CustomKeycloakProfile extends KeycloakProfile {
...
 }
 /**
  * Creates KeycloakApiService
  */
```
Hanhz2707, line 23:
```suggestion

/**
 * Creates KeycloakApiService
 */
```

Hanhz2707 (approved):
Minor changes, otherwise all good

src/functions/keycloak/remove-user-attribute/handler.ts
```diff
@@ -0,0 +1,38 @@
...
+      throw new Error("Missing path parameters: id or attributeName.");
+    }
+
+    const api = CreateKeycloakApiService();
+
+    await api.removeUserAttribute(id, attributeName);
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ message: `User attribute '${attributeName}' removed successfully.` }),
```
Hanhz2707: Maybe changing '${attributeName}' to "${attributeName}"

src/database/services/keycloak-api-service.ts
```diff
@@ -75,7 +78,110 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+      id: string,
+      attributes: Record<string, string[]> 
+    ): Promise<void> => {
+      const bodyContent = {
+        attributes: attributes,
+      };
+
+      try {
+        const response = await fetch(
+          `${baseUrl}/admin/realms/${realm}/users/${id}`,
```
dalju005-xamk, line 99: Can you call findUser() here to get user by id ?
Also same for other.

src/functions/keycloak/remove-user-attribute/handler.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+const removeUserAttributeHandler: APIGatewayProxyHandler = async (
+  event: APIGatewayProxyEvent
+) => {
+  try {
+    const id = event.pathParameters?.id;
+    const attributeName = event.pathParameters?.attributeName;
+
+    if (!id || !attributeName) {
+      throw new Error("Missing path parameters: id or attributeName.");
```
dalju005-xamk: Should this return some statuscode + body (errormessage)

src/functions/keycloak/update-user/handler.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Lambda handler to update a user's attributes
+ * 
+ * @param event API Gateway event 
+ * @returns Response message as JSON string
+ */
+const updateUserAttributeHandler: APIGatewayProxyHandler = async (
+  event: APIGatewayProxyEvent
+) => {
+  try {
+    if (!event.body) {
```
dalju005-xamk: Same as in other handler

dalju005-xamk (approved):

src/database/services/keycloak-api-service.ts
```diff
@@ -75,7 +78,110 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+        }
+
+      } catch (error) {
+        throw new Error(
+          error instanceof Error
+            ? error.message
+            : "An unknown error occurred while updating user attribute."
+        );
+      }
+    },
```
Jdallos, line 155:
```suggestion
    };
```

src/database/services/keycloak-api-service.ts
```diff
@@ -75,7 +78,110 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+     */
+    updateUserAttribute: async (
+      id: string,
+      attributes: Record<string, string[]> 
+    ): Promise<void> => {
+      const bodyContent = {
+        attributes: attributes,
+      };
+
+      try {
```
Jdallos, line 97: Looks good, however I now notice that the findUser method is missing the try catch handling, could you also set this up?

src/functions/keycloak/remove-user-attribute/handler.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @returns Response message as JSON string
+ */
+const removeUserAttributeHandler: APIGatewayProxyHandler = async (
+  event: APIGatewayProxyEvent
+) => {
+  try {
+    const id = event.pathParameters?.id;
+    const attributeName = event.pathParameters?.attributeName;
+
+    if (!id || !attributeName) {
```
Jdallos, line 16: I like that the removing and updating of the user attributes has been done in a generic way so it could update any attribute, but I'm also a little concerned it could be too much power when dealing with keycloak!
Perhaps we could add to this check that the attribute name is severaId attribute (whatever we decided to call it).
This could then easily be removed / extended in the future.

src/functions/keycloak/remove-user-attribute/handler.ts
```diff
@@ -0,0 +1,41 @@
...
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ message: "Missing required path parameters: 'id' or 'attributeName'." }),
```
Jdallos: Due to the above comment we might want to tweak the message e.g. or valid 'attributeName'

src/database/services/keycloak-api-service.ts
```diff
@@ -75,7 +78,110 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+        const response = await fetch(
```
Jdallos: I suppose as we are not returning from these services maybe typing is not essential. However please do some additional investigation about types from keycloak as it would be very nice to have a valid type for the fetch response from a given endpoint.

src/database/services/keycloak-api-service.ts
```diff
@@ -75,7 +78,110 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+  
+      if (!response.ok) {
+        const errorText = await response.text();
+        throw new Error(`Failed to fetch user data: ${response.status} - ${response.statusText}. Details: ${errorText}`);
+      }
+  
+      const user = await response.json();
+      const currentAttributes = user.attributes || {};
+  
+      delete currentAttributes[attributeName];
```
Jdallos, line 186: Not seen this syntax before, very nice

src/database/services/keycloak-api-service.ts
Jdallos: If you do find a way to type the fetch response, please update all fetch response types in the service

src/functions/keycloak/remove-user-attribute/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "delete",
+        path: "/users/{id}/attributes/{attributeName?}",
```
Jdallos: Also remember we typically design the new lambda endpoints in the home-lambdas-api-spec first. I haven't seen the spec updates yet so please make sure they get done.
In the spec you will describe the 2 endpoints which should match the path in the handler index files- so in this file for example. Adding in any relevant schemas (I guess none in this case as it doesn't return anything) and the path parameters and their types e.g. the type of attribute.
Also in the spec we should also describe that the only valid attribute name to be updated or removed is the severaId

src/functions/keycloak/update-user/handler.ts
```diff
@@ -0,0 +1,47 @@
...
+      throw new Error("Request body is missing.");
+    }
+
+    const body = JSON.parse(JSON.stringify(event.body));
+    const { id, attributes } = body;
+
+    if (!id || !attributes || typeof attributes !== "object") {
```
Jdallos: Again a similar check here on the key of the attribute to limit it to only updating if the key is severaId

src/functions/keycloak/remove-user-attribute/index.ts
Jdallos: I think {attributeName} is fine, so the ? can be removed or is it there for a reason?

src/functions/keycloak/update-user/handler.ts
```diff
@@ -0,0 +1,47 @@
...
+const updateUserAttributeHandler: APIGatewayProxyHandler = async (
+  event: APIGatewayProxyEvent
+) => {
+  try {
+    if (!event.body) {
+      throw new Error("Request body is missing.");
+    }
+
+    const body = JSON.parse(JSON.stringify(event.body));
+    const { id, attributes } = body;
```
Jdallos: Why not attribute rather than attributes? If for now it's single the name should be singular.

src/functions/keycloak/update-user/handler.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    if (!id || !attributes || typeof attributes !== "object") {
+      throw new Error("Missing or invalid body parameters: id, attributes.");
+    }
+
+    const api = CreateKeycloakApiService();
+
+    if (!attributes.isActive) {
+      attributes.isActive = ["Active"];
+    }
```
Jdallos: I'm not sure about this, shouldn't the attribute name be severaId and shouldn't the value be the severaId?
As such this lambda also seems to be missing the functionality to fetch the user from Severa to obtain the id and add it as an attribute. Perhaps the lambda can be actually be provided the severaId as part of the attribute.

I'm thinking when the user clicks the checkbox to grant permission in the UI, first there is a call to severa to update the users keyword, then the lambda returns the user Id, which is passed to this lambda call for updating the keycloak user. It might be also reasonable to combine all of this functionality into a single lambda making both severa and keycloak requests rather than splitting it apart.

In fact I think combining the functionality in these lambdas is the way to go, but let me know what you think is best. Also something similar needs to be done for the remove e.g. not just remove the attribute from keycloak but also removing the keyword from the severa user.

It might be worth checking with Ilmo incase anyone else has started onthe severa Keyword functionality but I'm not aware of it,

src/functions/keycloak/remove-user-attribute/handler.ts
```diff
@@ -0,0 +1,41 @@
...
+      };
+    }
+
+    const api = CreateKeycloakApiService();
+
+    await api.removeUserAttribute(id, attributeName);
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify({ message: `User attribute "${attributeName}" removed successfully.` }),
```
Jdallos: If we are treating this as a delete endpoint it should return the appropriate status and body e,g,
   return {
      statusCode: 204,
      body: "",
    };

src/functions/keycloak/update-user/handler.ts
Jdallos, line 1: Shouldn't the file be name update-user-attributes

src/functions/keycloak/update-user/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "put",
+        path: "/users/{id}",
```
Jdallos: Shouldn't this path also be:
/users/{id}/attributes/{attributeName}

??

Jdallos (changes requested):

src/functions/keycloak/update-user/index.ts
rimz178: 'll keep the original for now because I think it works better this way.

serverless.ts
```diff
@@ -80,6 +82,11 @@ const serverlessConfiguration: AWS = {
     environment: {
       AWS_NODEJS_CONNECTION_REUSE_ENABLED: "1",
       NODE_OPTIONS: "--enable-source-maps --stack-trace-limit=1000",
+      NODE_ENV: "development",
+      SEVERA_TEST_USER_EMAIL: env.SEVERA_TEST_USER_EMAIL,
+      SEVERA_TEST_USER_ID: env.SEVERA_TEST_USER_ID,
+
+
```
Hanhz2707: 2 empty line here

serverless.ts
```diff
@@ -115,11 +122,12 @@ const serverlessConfiguration: AWS = {
       TRELLO_MANAGEMENT_BOARD_ID: env.TRELLO_MANAGEMENT_BOARD_ID,
       CHANNEL_ID: env.CHANNEL_ID,
       OPENAI_API_KEY: env.OPENAI_API_KEY,
+
```
Hanhz2707: empty line

src/database/services/keycloak-api-service.ts
```diff
@@ -74,8 +76,148 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+    /**
+     * Updates a user's attributes
+     * 
+     * @param id string
+     * @param attributes  Record<string, string[]>
+     */
+    updateUserAttribute: async (
+      id: string,
+      attributes: Record<string, string[]>
+
```
Hanhz2707: empty line

src/database/services/keycloak-api-service.ts
```diff
@@ -74,8 +76,148 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+        const updatedAttributes = {
+          ...existingAttributes,
+          ...attributes,
+        };
+
+        const bodyContent = {
+          attributes: updatedAttributes,
+          email: existingEmail, 
+        };
+        console.log("Updating user with attributes:", attributes);
```
Hanhz2707: Do we need this in develop ?

src/database/services/keycloak-api-service.ts
```diff
@@ -74,8 +76,148 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+          error instanceof Error
+            ? error.message
+            : "An unknown error occurred while updating user attribute"
+        );
+      }
     },
-  };
+    
+  /**
+   * removes user attributes
```
Hanhz2707:
```suggestion
   * Remove user attributes
```

src/database/services/keycloak-api-service.ts
```diff
@@ -74,8 +76,148 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+          }
+        );
+
+        if (!updateResponse.ok) {
+          const errorText = await updateResponse.text();
+          throw new Error(
+            `Failed to update user attributes: ${updateResponse.status} - ${updateResponse.statusText}. Details: ${errorText}`
+          );
+        }
+        console.log("User attributes updated successfully.");
```
Hanhz2707: Do we need this ?

src/database/services/keycloak-api-service.ts
```diff
@@ -74,8 +76,148 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+      } catch (error) {
+        console.error("Error updating user attributes:", error);
```
Hanhz2707: Do we need this ?

src/database/services/keycloak-api-service.ts
```diff
@@ -74,8 +76,148 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+      }
+  
+      const user = await response.json();
+      const existingEmail = user.email;
+
+      const currentAttributes = user.attributes || {};
+
+      delete currentAttributes[attributeName];
+  
+
```
Hanhz2707: Empty line

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,89 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import { middyfy } from "src/libs/lambda";
+import { CreateSeveraApiService } from "src/services/severa-api-service";
+
+
```
Hanhz2707, line 5: Empty line

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,89 @@
...
+    };
+  }
+
+  const api = CreateKeycloakApiService();
+  const severaApi = CreateSeveraApiService()
+
+  let severaUser: {email:string, guid:string } | null = null
+
+// If in development environment, use the test user defined in environment variables
+// To make this work, add SEVERA_TEST_USER_EMAIL and SEVERA_TEST_USER_ID to your .env file
```
Hanhz2707: delete this comment, and self - check all others console.log within this file, do they needed in develop ?

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  if (!attributes.isActive) {
+    attributes.isActive = ["Active"];
+  }
+  attributes["severa-user-id"] = [severaUser.guid];
+  console.log("atrributes info", attributes )
+  await api.updateUserAttribute(id, attributes);
+
+
```
Hanhz2707: empty line

src/database/services/keycloak-api-service.ts
naStipe: this should not be here

src/database/services/keycloak-api-service.ts
naStipe: This also should not be here

src/database/services/keycloak-api-service.ts
naStipe: I am not sure if we want to use this kind of errors

src/services/severa-api-service.ts
```diff
@@ -21,6 +21,8 @@ export interface SeveraApiService {
   getWorkDays: (severaUserId: string) => Promise<SeveraResponseWorkDays>;
   getOptInUsers: () => Promise<SeveraResponseUser[]>;
   getResourceAllocations: () => Promise<SeveraResponseResourceAllocation>;
+  getUserByEmail: (email: string, attribute: Record<string, string[]>) => Promise<{ guid: string; isSeveraOptIn:string, email:string }> ;
+  getTestUser:()  => Promise<{ guid: string, email: string }>
```
Hanhz2707:
```suggestion
  getTestUser:() => Promise<{ guid: string, email: string }>
```

Hanhz2707: Extra space

src/database/services/keycloak-api-service.ts
```diff
@@ -74,8 +76,148 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+          const errorText = await updateResponse.text();
+          throw new Error(
+            `Failed to update user attributes: ${updateResponse.status} - ${updateResponse.statusText}. Details: ${errorText}`
+          );
+        }
+        console.log("User attributes updated successfully.");
+      } catch (error) {
+        console.error("Error updating user attributes:", error);
+        throw new Error(
+          error instanceof Error
```
naStipe, line 150: I feel like this is unnecessary, cuz I haven`t seen us using it this way anywhere else

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,156 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        const response = await fetch(url, {
+          method: "GET",
+          headers: {
+            Authorization: `Bearer ${await getSeveraAccessToken()}`,
+            "Client_Id": process.env.SEVERA_DEMO_CLIENT_ID,
+            "Content-Type": "application/json",
+          },
+        });
+    
+        if (!response.ok) {
```
Hanhz2707: Please also check all the console.log here, are they needed here ?

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,156 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+          });
+    
+          if (!allKeywordsResponse.ok) {
+            throw new Error(`Failed to fetch all keywords: ${allKeywordsResponse.status} - ${allKeywordsResponse.statusText}`);
+          }
+    
+          const allKeywords = await allKeywordsResponse.json();
+          let keywordToAdd = allKeywords.find((kw: { keyword: string }) => kw.keyword === "isSeveraOptIn");
+
+          if (!keywordToAdd) {
```
Hanhz2707: Empty line

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,156 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+              throw new Error(`Failed to update Severa keyword: ${updateResponse.status} - ${updateResponse.statusText}`);
+            }
+    
+            console.log(`Updated keyword 'isSeveraOptIn' successfully for user ${user.guid}`);
+          }
+        }
+        return {
+          guid: user.guid,
+          isSeveraOptIn, 
+          email,
```
Hanhz2707: Empty line

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,89 @@
...
+  const api = CreateKeycloakApiService();
+  const severaApi = CreateSeveraApiService()
+
+  let severaUser: {email:string, guid:string } | null = null
+
+// If in development environment, use the test user defined in environment variables
+// To make this work, add SEVERA_TEST_USER_EMAIL and SEVERA_TEST_USER_ID to your .env file
+  if (process.env.NODE_ENV === "development") {
+    console.log("Current NODE_ENV:", process.env.NODE_ENV);
```
naStipe: Please, remove all console logs from the project

Hanhz2707 (changes requested):
There are suggestions

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  const api = CreateKeycloakApiService();
+  const severaApi = CreateSeveraApiService()
+
+  let severaUser: {email:string, guid:string } | null = null
+
+// If in development environment, use the test user defined in environment variables
+// To make this work, add SEVERA_TEST_USER_EMAIL and SEVERA_TEST_USER_ID to your .env file
+  if (process.env.NODE_ENV === "development") {
```
naStipe: I saw that this is defined in "serverless.ts", should we define it in envs?

naStipe (approved):

src/functions/keycloak/update-user-attributes/handler.ts
rimz178: Yes, we should add the information there in the envs.

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,82 @@
...
+      statusCode: 400,
+      body: JSON.stringify({ message: "Attributes contain invalid keys." }),
+    };
+  }
+
+  const api = CreateKeycloakApiService();
+  const severaApi = CreateSeveraApiService()
+
+  let severaUser: {email:string, guid:string } | null = null
+
```
Hanhz2707, line 48: empty line

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,82 @@
...
+    body: JSON.stringify({ message: "IsSeveraOptin has already been updated in Severa, and severaId will be added to Keycloak." }),
+  };
+} catch (error) {
+  return {
+    statusCode: 500,
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({ message: error.message, stack: error.stack }),
+  };
+}
+
```
Hanhz2707: empty line

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,153 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        }
+    
+        return {
+        email: process.env.SEVERA_TEST_USER_EMAIL,
+        guid: process.env.SEVERA_TEST_USER_ID 
+        };
+      } catch (error) {
+        console.error("Error in getTestUser:", error);
+        throw error;
+      }
```
Hanhz2707: I thought we already has throw error here , why we also need console.error ?

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,153 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+    },
+
+    /**
+     * Fetches a Severa user by their email address and checks/updates the 'isSeveraOptIn' status.
+    * 
+    * @param email The email address of the user to be retrieved.
+    * @param attribute Attributes to be checked/updated for the user.
+    * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
+     */
```
Hanhz2707: Fix indentation

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,153 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        guid: process.env.SEVERA_TEST_USER_ID 
+        };
+      } catch (error) {
+        console.error("Error in getTestUser:", error);
+        throw error;
+      }
+    },
+
+    /**
+     * Fetches a Severa user by their email address and checks/updates the 'isSeveraOptIn' status.
```
Hanhz2707: Indentation

src/services/severa-api-service.ts
Hanhz2707:
```suggestion
    /**
     * Fetches a Severa user by their email address and checks/updates the 'isSeveraOptIn' status.
     * 
     * @param email The email address of the user to be retrieved.
     * @param attribute Attributes to be checked/updated for the user.
     * 
     * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
     */

File: src/services/severa-api-service.ts [hunk 43]
```diff
@@ -57,6 +59,153 @@ export const CreateSeveraApiService = (): SeveraApiService => {
       }
       return response.json();
     },
+      
+    /**
```
Hanhz2707: Fix indentation

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,153 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+            if (!updateResponse.ok) {
+              throw new Error(`Failed to update Severa keyword: ${updateResponse.status} - ${updateResponse.statusText}`);
+            }
+          }
+        }
+        return {
+          guid: user.guid,
+          isSeveraOptIn, 
+          email,
+
```
Hanhz2707: Empty line

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,153 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        };
+    
+      } catch (error) {
+
```
Hanhz2707: Empty line

src/types/severa/user/severaResponseUser.ts
```diff
@@ -6,6 +6,7 @@ interface SeveraResponseUser {
   name: string;
   firstName: string;
   lastName: string;
+  email: string,
```
Hanhz2707:
```suggestion
  email: string;
```

Hanhz2707 (approved):

serverless.ts
```diff
@@ -80,6 +82,9 @@ const serverlessConfiguration: AWS = {
     environment: {
       AWS_NODEJS_CONNECTION_REUSE_ENABLED: "1",
       NODE_OPTIONS: "--enable-source-maps --stack-trace-limit=1000",
+      NODE_ENV: "development",
```
Jdallos: This is hardcoded, meaning it won't work in the productino environment. I don't think you need this here.
We have the isLocal variable in the serverless.ts which is based off process.end.STAGE- could you use this?

```suggestion
```

src/database/services/keycloak-api-service.ts
```diff
@@ -55,7 +57,7 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
      * @returns user by Id
      */
     findUser: async (id: string): Promise<CustomKeycloakProfile> => {
-      const response = await fetch(
+      try { const response = await fetch(
```
Jdallos:
```suggestion
      try { 
      const response = await fetch(
```

Jdallos: Line break when opening try block

src/functions/keycloak/remove-user-attribute/handler.ts
```diff
@@ -0,0 +1,48 @@
...
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import { middyfy } from "src/libs/lambda";
+
+/**
+ * Lambda handler to remove a user's attribute
+ * 
+ * @param event API Gateway event
+ * @returns Response message as JSON string
+ */
+const removeUserAttributeHandler: APIGatewayProxyHandler = async (
```
Jdallos: As I sent in a message the other week I think we might not actually need the removal function, but there is no need for us to remove the code but probably we won't call it in the UI for now.

src/functions/keycloak/remove-user-attribute/handler.ts
```diff
@@ -0,0 +1,48 @@
...
+    const attributeName = event.pathParameters?.attributeName;
+
+    if (!id || !attributeName) {
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ message: "Missing required path parameters: 'id' or  valid 'attributeName'." }),
+      };
+    }
+    
+    if (attributeName !== 'isSeveraOptIn') {
```
Jdallos:
```suggestion
    if (attributeName !== "isSeveraOptIn") {
```

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Lambda handler to update a user's attributes
+ * 
+ * @param event API Gateway event 
+ * @returns Response message as JSON string
+ */
+const updateUserAttributeHandler: APIGatewayProxyHandler = async (
+  event: APIGatewayProxyEvent
+  ) => { 
+  try {
+  if (!event.body) {
```
Jdallos: Indentation needs fixing within the try block, check indentation throughout this file

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+  const api = CreateKeycloakApiService();
+  const severaApi = CreateSeveraApiService()
+
+  let severaUser: {email:string, guid:string } | null = null
+
+  if (process.env.NODE_ENV === "development") {
+    severaUser = await severaApi.getTestUser();
+  } else {
+    severaUser = await severaApi.getUserByEmail(email, attributes);
+  }
```
Jdallos: Rather than have the condition here and have 2 functions, why not just have the NODE_ENV (which should probably be updated to process.env.STAGE, as said in a previous comment) check within the getUserByEmail function, if in local then it should return the appropriate envs. 
Also we should not need to do the checks to make sure the envs are available, rather you should add any new envs into the sec/app/config.ts

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+    return {
+      statusCode: 400,
+      body: JSON.stringify({ message: "Attributes contain invalid keys." }),
+    };
+  }
+
+  const api = CreateKeycloakApiService();
+  const severaApi = CreateSeveraApiService()
+
+  let severaUser: {email:string, guid:string } | null = null
```
Jdallos:
```suggestion
  let severaUser: { email:string, guid:string } | null = null
```

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  if (process.env.NODE_ENV === "development") {
+    severaUser = await severaApi.getTestUser();
+  } else {
+    severaUser = await severaApi.getUserByEmail(email, attributes);
+  }
+  if (!severaUser || !severaUser.email) {
```
Jdallos:
```suggestion
  if (!severaUser?.email || !severaUser?.guid) {
```

Jdallos: Is this suggestion better?

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+    return {
+      statusCode: 404,
+      body: JSON.stringify({ message: "Severa user not found." }),
+    };
+  }
+
+  if (!attributes.isActive) {
+    attributes.isActive = ["Active"];
+  }
```
Jdallos: Can you explain what this is for and why we need it?

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+  attributes["severa-user-id"] = [severaUser.guid];
```
Jdallos: Looks to me like in keycloak we have sever-user-id and severaUserId as attributes. As most attributes in keycloak are in camelCase lets use the camel case version here also.

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+            "Content-Type": "application/json",
+          },
+        });
+    
+        if (!response.ok) {
+          throw new Error(`Failed to fetch user by email: ${response.status} - ${response.statusText}`);
+        }
+    
+        const users = await response.json();
+        if (!users || users.length === 0) {
```
Jdallos:
```suggestion
        if (!users?.length) {
```

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+      email: process.env.SEVERA_TEST_USER_EMAIL,
+      guid: process.env.SEVERA_TEST_USER_ID 
+      };
+    },
+
+    /**
+    * Fetches a Severa user by their email address and checks/updates the 'isSeveraOptIn' status.
+    * 
+    * @param email The email address of the user to be retrieved.
+    * @param attribute Attributes to be checked/updated for the user.
```
Jdallos: In keycloak we are working with attributes, but in Severa we are updating the keyword and that's how we should be referring to it otherwise it starts to get confusing. Update param name and all other references to attribute when you are in face talking about keyword

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+    
+        if (!keywordsResponse.ok) {
+          throw new Error(`Failed to fetch keywords: ${keywordsResponse.status} - ${keywordsResponse.statusText}`);
+        }
+    
+        const keywords = await keywordsResponse.json();
+        const isSeveraOptInKeyword = keywords.find((kw: { keyword: string }) => kw.keyword === "isSeveraOptIn");
+    
+        if (isSeveraOptInKeyword) {
+
```
Jdallos: Why is this block empty, if it's not needed it should be removed. In which case the else can be if (!isSeveraOptInKeyword)

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        const users = await response.json();
+        if (!users || users.length === 0) {
+          throw new Error(`No user found with email: ${email}`);
+        }
+    
+        const user = users[0];
+    
+        const isSeveraOptIn = attribute.isSeveraOptIn?.[0]; 
+    
+        const keywordsUrl = `${baseUrl}/v1/users/${user.guid}/keywords`;
```
Jdallos: This could be named userKeywordsUrl

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+          throw new Error(`Failed to fetch keywords: ${keywordsResponse.status} - ${keywordsResponse.statusText}`);
+        }
+    
+        const keywords = await keywordsResponse.json();
+        const isSeveraOptInKeyword = keywords.find((kw: { keyword: string }) => kw.keyword === "isSeveraOptIn");
+    
+        if (isSeveraOptInKeyword) {
+      
+        } else {
+          const allKeywordsUrl = `${baseUrl}/v1/keywords?keyword=isSeveraOptIn`;
```
Jdallos: This could be named keywordsUrl

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+                isActive: true,
+              }),
+            });
+    
+            if (!createKeywordResponse.ok) {
+              throw new Error(`Failed to create 'isSeveraOptIn' keyword: ${createKeywordResponse.status} - ${createKeywordResponse.statusText}`);
+            }
+    
+            keywordToAdd = await createKeywordResponse.json();
+          }
```
Jdallos: Don't we only need to add the keyword to severa once?
This just seems to bulk out the function quite a bit for something that should only really need to be done once on set up e.g. it could be done in postman? Perhaps we extract this into a checkKeywordExists function and call it here rather than have all the code here. Assuming the keyword does exist (it should after the first time) we then proceed to the if (keywordToAdd) condition, which I would suggest renaming to reflect that the keyword exists in severa and does not exist on a user.

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+            });
+    
+            if (!updateResponse.ok) {
+              throw new Error(`Failed to update Severa keyword: ${updateResponse.status} - ${updateResponse.statusText}`);
+            }
+          }
+        }
+        return {
+          guid: user.guid,
+          isSeveraOptIn,
```
Jdallos: rather than returning the isSeveraOptIn could we return something from theUpdateKeyword response? Perhaps still the keyword, but it's better coming from the successful response rather than essentially from the attributes provided to the function in the first place.

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+          }
+        }
+        return {
+          guid: user.guid,
+          isSeveraOptIn, 
+          email,
+        };
+    
+      } catch (error) {
+        throw new Error(error instanceof Error ? error.message : "An unknown error occurred.");
```
Jdallos: Can we give more details in the case of unknown error occured here, e.g. while finding user- although this function is doing quite a bit more than that

src/services/severa-api-service.ts
```diff
@@ -57,6 +59,146 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+
+    /**
+    * Fetches a Severa user by their email address and checks/updates the 'isSeveraOptIn' status.
+    * 
+    * @param email The email address of the user to be retrieved.
+    * @param attribute Attributes to be checked/updated for the user.
+    * 
+    * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
+    */
+    getUserByEmail: async (email: string, attribute: Record<string, string[]>) => {
```
Jdallos: Typically in the service we would break down each function to call one API endpoint and then call the various endpoints we need from the lambda handler. In this case you are getting user by email, you are also listing keywords for a user, potentially listing all keywords, potentially adding a new keyword and then updating a user keyword.

So getUserByEmail is not an appropriate name and I would break this function down into 5 smaller functions within the service, each of which can then be  called from the handler. 

Hopefully this makes sense but happy to discuss/ clarify if needed. Currently this fuction seems to work but needs to be refactored- easier to handle errors, easier to read/ debug etc.

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  await api.updateUserAttribute(id, attributes);
+
+  await severaApi.getUserByEmail(severaUser.email, attributes);
+  
+  return {
+    statusCode: 200,
+    body: JSON.stringify({ message: "IsSeveraOptin has already been updated in Severa, and severaId will be added to Keycloak." }),
```
Jdallos: I've left several comments in the getUserByEmail function which needs refactoring. but here you say the IsSeveraOptin (should be OptIn) and it is, but this is not at all clear from teh function name getUserByEmail. ALso why is severa in past tense but you say severaId WIll be added to keycloak.

This handler response should be a message: "Severa keyword updated: {data1} , keycloak attribute updated: {data2}"

You should pass any relevant data from the API responses once the keyword or attribute have been updated.

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -0,0 +1,80 @@
...
+  };
+} catch (error) {
+  return {
+    statusCode: 500,
+    headers: { "Content-Type": "application/json" },
+    body: JSON.stringify({ message: error.message, stack: error.stack }),
+  };
+}
+};
```
Jdallos: Indentation

src/types/severa/user/severaResponseUser.ts
```diff
@@ -6,6 +6,7 @@ interface SeveraResponseUser {
   name: string;
   firstName: string;
   lastName: string;
+  email: string;
```
Jdallos: Can you explain why you have added this as it seems you are not using it anywhere in your updated code

Jdallos (changes requested):

src/functions/keycloak/update-user-attributes/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "put",
+        path: "/users/{id}",
```
Jdallos: Shouldn't this be "/users/{id}/attributes/{attributeName}"?

src/types/severa/user/severaResponseUser.ts
rimz178: I seem to have forgotten to remove this :D

src/functions/keycloak/update-user-attributes/handler.ts
rimz178: IsActive is here for this reason, because if I update IsSeveraOptIn, then IsActive always has to be included either in the body or in the code. I'm not sure if IsActive needs to be in Keycloak, but Stephan mentioned that it should always be included when updating the attributes

rimz178: IsActive is included here because if I update IsSeveraOption, IsActive always needs to be included in the body or in the code. I'm not entirely sure if IsActive is necessary in Keycloak, but Stephan mentioned that it should always be included when updating the attributes.

src/functions/keycloak/update-user-attributes/handler.ts
rimz178: it didn't allow me to update Keycloak with that severeUserId

src/app/config.ts
```diff
@@ -35,6 +37,11 @@ export default class Config {
       apiId: env.SPLUNK_API_ID,
       teamOnCallUrl: env.SPLUNK_TEAM_ONCALL_URL,
       schedulePolicyName: env.SPLUNK_SCHEDULE_POLICY_NAME
+    },
+    testUser: {
+      email: env.SEVERA_TEST_USER_EMAIL,
+      id: env.SEVERA_TEST_USER_ID,
     }
+
```
Hanhz2707: Empty line

src/database/services/keycloak-api-service.ts
```diff
@@ -1,6 +1,5 @@
 import fetch from "node-fetch";
 import type { KeycloakProfile } from "keycloak-js/lib/keycloak";
-
 /**
```
Hanhz2707, line 3:
```suggestion
/**
```
import type { KeycloakProfile } from "keycloak-js/lib/keycloak";

/**
 * Custom Interface for a user in keycloak functions with severaUserId added.
 */

Hanhz2707:
```suggestion
import fetch from "node-fetch";
import type { KeycloakProfile } from "keycloak-js/lib/keycloak";

/**
 * Custom Interface for a user in keycloak functions with severaUserId added.
 */
```

Hanhz2707: You can give them an extra empty line between 2 different interfaces in line 8,9

src/database/services/keycloak-api-service.ts
```diff
@@ -74,8 +77,145 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+          }
+        );
+
+        if (!updateResponse.ok) {
+          const errorText = await updateResponse.text();
+          throw new Error(
+            `Failed to update user attributes: ${updateResponse.status} - ${updateResponse.statusText}. Details: ${errorText}`
+          );
+        }
+
```
Hanhz2707:
```suggestion
        if (!updateResponse.ok) {
          const errorText = await updateResponse.text();
          throw new Error(
            `Failed to update user attributes: ${updateResponse.status} - ${updateResponse.statusText}. Details: ${errorText}`
          );
        }
      } catch (error) {
```

src/functions/keycloak/remove-user-attribute/index.ts
```diff
@@ -0,0 +1,16 @@
...
+      httpApi: {
+        method: "delete",
+        path: "/users/{id}/attributes/{attributeName}",
+        authorizer: {
+            name: "timebankKeycloakAuthorizer",
+          },
+        },
+      },
+    ],
+};
```
Hanhz2707:
```suggestion
  events: [
    {
      httpApi: {
        method: "delete",
        path: "/users/{id}/attributes/{attributeName}",
        authorizer: {
          name: "timebankKeycloakAuthorizer",
        },
      },
    },
  ],
};
```

src/types/index.ts
```diff
@@ -26,4 +26,8 @@ export interface Configuration {
         teamOnCallUrl: string,
         schedulePolicyName: string
     }
+    testUser: {
+        email: string,
+        id: string
+    }
 }
```
Hanhz2707, line 33: I just noticed this file is having space 4. Could you change to spaces:2 for this file. Thanks!

naStipe (approved):

src/database/services/keycloak-api-service.ts
```diff
@@ -57,7 +59,8 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
      * @returns user by Id
      */
     findUser: async (id: string): Promise<CustomKeycloakProfile> => {
-      try { const response = await fetch(
+      try { 
+      const response = await fetch(
```
Jdallos: Fix indentation

src/functions/keycloak/remove-user-attribute/index.ts
```diff
@@ -8,7 +8,7 @@ export default {
         method: "delete",
         path: "/users/{id}/attributes/{attributeName}",
         authorizer: {
-            name: "timebankKeycloakAuthorizer",
+          name: "timebankKeycloakAuthorizer",
           },
         },
       },
```
Jdallos: Fix indentation

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -5,76 +5,78 @@ import { CreateSeveraApiService } from "src/services/severa-api-service";
...
-  const severaApi = CreateSeveraApiService()
+    if (!Object.prototype.hasOwnProperty.call(attributes, attributeName)) {
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ message: `Attribute ${attributeName} is missing in request body.` })
+      };
+    }
 
-  let severaUser: {email:string, guid:string } | null = null
+    const severaUser = await severaApi.getUserByEmail(email, attributes);
```
Jdallos: I'm suspicious about the name, if this method is just getting user by email, why does it need attributes?

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -5,76 +5,78 @@ import { CreateSeveraApiService } from "src/services/severa-api-service";
...
+    }
 
-  const keys = Object.keys(attributes);
-  if (!keys.every(key => allowedKeys.includes(key))) {
-    return {
-      statusCode: 400,
-      body: JSON.stringify({ message: "Attributes contain invalid keys." }),
-    };
-  }
+    const api = CreateKeycloakApiService();
```
Jdallos, line 38: could call it keycloakApi?

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -5,76 +5,78 @@ import { CreateSeveraApiService } from "src/services/severa-api-service";
...
-    headers: { "Content-Type": "application/json" },
-    body: JSON.stringify({ message: error.message, stack: error.stack }),
-  };
-}
+    await severaApi.getUserByEmail(severaUser.email, attributes);
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify({
+        message: `Severa user updated with: ${JSON.stringify({ attributes })}, Keycloak attributes updated: ${JSON.stringify(attributes)}`
```
Jdallos: As I said in a previous comment don't we get any kind of return values form the API's that we can return here, rather than just returning the attributes that the user provides. 
I think if there are bugs/ anything weird going on (now or in the future with further development) doing this will make them easier to spot

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -5,76 +5,78 @@ import { CreateSeveraApiService } from "src/services/severa-api-service";
...
-    throw new Error("Request body is missing.");
-  }
+    if (!event.body) {
+      throw new Error("Request body is missing.");
+    }
 
-  const body = JSON.parse(JSON.stringify(event.body));
-  const { id, attributes } = body;
-  const email = attributes?.email; 
+    const body = JSON.parse(JSON.stringify(event.body));
```
Jdallos, line 18: Are you sure this won't work:
const body = JSON.parse(event.body);

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -5,76 +5,78 @@ import { CreateSeveraApiService } from "src/services/severa-api-service";
...
-    body: JSON.stringify({ message: "IsSeveraOptin has already been updated in Severa, and severaId will be added to Keycloak." }),
-  };
-} catch (error) {
-  return {
-    statusCode: 500,
-    headers: { "Content-Type": "application/json" },
-    body: JSON.stringify({ message: error.message, stack: error.stack }),
-  };
-}
+    await severaApi.getUserByEmail(severaUser.email, attributes);
```
Jdallos: IT's odd you are calling this method again. Did you not break this don't into smaller keycloak methods as I suggested. E.g. one to find a user based on email, one to update the keyword etc. Rather than having them all in the the same method with a somewhat misleading name

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -5,76 +5,78 @@ import { CreateSeveraApiService } from "src/services/severa-api-service";
...
+    const severaApi = CreateSeveraApiService();
 
-  const api = CreateKeycloakApiService();
-  const severaApi = CreateSeveraApiService()
+    if (!Object.prototype.hasOwnProperty.call(attributes, attributeName)) {
+      return {
+        statusCode: 400,
+        body: JSON.stringify({ message: `Attribute ${attributeName} is missing in request body.` })
+      };
+    }
```
Jdallos: I'm pretty sure this is redundant and can be removed as you have this earlier code which is checking the attributes are allowed key:
 const keys = Object.keys(attributes);
    if (!keys.every((key) => allowedKeys.includes(key))) {
      return {
        statusCode: 400,
        body: JSON.stringify({ message: "Attributes contain invalid keys." })
      };
    }

src/services/severa-api-service.ts
```diff
@@ -33,200 +34,113 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+     * @param email The email address of the user to be retrieved.
+     * @param keyword Keywords to be checked/updated for the user.
+     *
+     * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
+     */
+    getUserByEmail: async (email: string, keyword: Record<string, string[]>) => {
+      dotenv.config();
+      const isLocal = process.env.STAGE === "local";
+
+      const testEmail = isLocal ? Config.get().testUser.email : email;
```
Jdallos: If this is handling the develop environemnt "test email" and the actual user provided email as it seems to be, then this should be renamed to email, or as we already have the email param, i guess, "userEmail"

src/services/severa-api-service.ts
```diff
@@ -33,200 +34,113 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
-    */
-    getUserByEmail: async (email: string, attribute: Record<string, string[]>) => {
+     * Fetches a Severa user by their email address and checks/updates the 'isSeveraOptIn' status.
+     *
+     * @param email The email address of the user to be retrieved.
+     * @param keyword Keywords to be checked/updated for the user.
+     *
+     * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
+     */
+    getUserByEmail: async (email: string, keyword: Record<string, string[]>) => {
```
Jdallos: So you have broken down the keycloak methods into separate functions which is great. This is like the higher level function calling the various methods to basically find user, update user, potentially create keyword if needed etc. So it should be renamed to something else more accurate, potentially- "optInSeveraUser"? 

Also 
I don't think this function needs to be in the service as it's combining various methods of the service and applying logic. I'd suggest creating a src/utils/severa/index.ts and moving this function in there. (If this seems to cause a lot of problems let me know but it should be simple to move).

src/services/severa-api-service.ts
```diff
@@ -33,200 +34,113 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+            email: testEmail
+          };
         }
         return {
           guid: user.guid,
-          isSeveraOptIn, 
-          email,
+          isSeveraOptIn: existingKeywordForUser.value,
+          email: testEmail,
+          testUserId: testUserId
```
Jdallos: Why is this testUserId needed here, at this point won't it be the same as user.guid?

src/services/severa-api-service.ts
```diff
@@ -33,200 +34,113 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+     * @param keyword Keywords to be checked/updated for the user.
+     *
+     * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
+     */
+    getUserByEmail: async (email: string, keyword: Record<string, string[]>) => {
+      dotenv.config();
+      const isLocal = process.env.STAGE === "local";
+
+      const testEmail = isLocal ? Config.get().testUser.email : email;
+      const testUserId = isLocal ? Config.get().testUser.id : null;
```
Jdallos: Is this actually needed or can it be removed?

src/services/severa-api-service.ts
Jdallos: You are returning from here though, which means you can return these in the body of the handler return- doesn't that make more sense?

src/functions/keycloak/update-user-attributes/handler.ts
Jdallos: So now I have gone through the getUserByEmail method and left comments related to that, but can you explain why you are calling this twice in the handler? You should only need to call it once.

Jdallos (changes requested):

src/functions/keycloak/update-user-attributes/handler.ts
rimz178: If I use that like this, I get the following error message: "[object Object]" is not valid JSON.

src/functions/keycloak/update-user-attributes/handler.ts
rimz178: Yeah, that's useless :D

src/functions/keycloak/update-user-attributes/handler.ts
rimz178: I get an error if I change it to that name.

src/functions/keycloak/update-user-attributes/handler.ts
rimz178: The getUserByEmail method requires both email and attributes parameters. Even though it seems like only the email is needed, the attributes parameter is necessary for the API to function correctly. If you remove it, the method might not work as expected. You can pass an empty object {} as the attributes to satisfy the method's requirements without affecting the result.

src/services/severa-api-service.ts
```diff
@@ -18,12 +16,11 @@ export interface SeveraApiService {
   getFlextimeBySeveraUserId: (severaUserId: string, eventDate: string) => Promise<Flextime>;
   getResourceAllocation: (endpointPath: URL) => Promise<SeveraResponseResourceAllocation[]>;
   getPhasesBySeveraProjectId: (severaProjectId: string) => Promise<SeveraResponsePhases[]>;
-  getWorkHours: (endpointPath: URL, startDate?: string,endDate?: string) => Promise<SeveraResponseWorkHours[]>;
+  getWorkHours: (endpointPath: URL,startDate?: string,endDate?: string ) => Promise<SeveraResponseWorkHours[]>;
```
Jdallos:
```suggestion
  getWorkHours: (endpointPath: URL, startDate?: string, endDate?: string ) => Promise<SeveraResponseWorkHours[]>;
```

src/utils/severa/index.ts
```diff
@@ -0,0 +1,57 @@
+import dotenv from "dotenv";
```
Jdallos: I'm not sure why you need this, you are envalid in the config to validate the testUserr envs, so can't you just get the envs from there and remove the dotenv references here?
Remove and confirm it still works

src/utils/severa/index.ts
```diff
@@ -0,0 +1,57 @@
...
+      guid: user.guid,
+      isSeveraOptIn: existingKeywordForUser.value,
+      email: userEmail
+    };
+  } catch (error) {
+    const errorMessage =
+      error instanceof Error
+        ? error.message
+        : "An unknown error occurred while processing the user request.";
+    throw new Error(`Error in getUserByEmail: ${errorMessage}`);
```
Jdallos:
```suggestion
    throw new Error(`Error while updating severa opt in: ${errorMessage}`);
```

src/functions/keycloak/update-user-attributes/handler.ts
```diff
@@ -60,14 +51,15 @@ const updateUserAttributeHandler: APIGatewayProxyHandler = async (event: APIGate
...
+    const updateResponse = await api.updateUserAttribute(id, attributes);
 
     return {
       statusCode: 200,
       body: JSON.stringify({
-        message: `Severa user updated with: ${JSON.stringify({ attributes })}, Keycloak attributes updated: ${JSON.stringify(attributes)}`
+        message: "Severa user and Keycloak attributes updated",
+        severaUser: severaUser,
+        updatedAttributes: attributes,
+        keycloakResponse: updateResponse
```
Jdallos, line 62: This seems better to me, just remember we need to describe what is being returned in the spec

Jdallos (approved):

PR #139, opened 2024-12-09
MetaAssistant daily message restoration

McDuuck:

PR #156, opened 2025-02-10
Feature 147 ma weekly message restoration

### LINKED ISSUE #147: MA Weekly message restoration

McDuuck:
This still need some work, This pr is a good reference:
https://github.com/Metatavu/home-lambdas/pull/139/files
When testing remember to add your own slackid to the .env file, and to test it you need to invoke the script.
Some data might be wrong but that should be clear when testing and some endpoints or parsing the data might need adjustment.
Also remember to deal with conflicts and keep this branch updated with develop.
Have fun.


src/functions/meta-assistant/send-weekly-message/handler.ts
```diff
@@ -1,93 +1,96 @@
...
-//     const timeRegistrations = await ForecastApiUtilities.getTimeRegistrations(weekStartDate);
-//     const nonProjectTimes = await ForecastApiUtilities.getNonProjectTime();
+    if (!severaUsers) {
+      throw new Error("No users retrieved from Severa");
+    }
 
-//     if (!timebankUsers) {
-//       throw new Error("No persons retrieved from Timebank");
-//     }
+    const personTotalTimes: WeeklyCombinedData[] = [];
```
naStipe: probably should be userTotalTimes?

src/functions/meta-assistant/send-weekly-message/handler.ts
```diff
@@ -1,93 +1,96 @@
...
-//     }
+      let totalWorkHours = 0;
+      let totalExpectedHours = 0;
+      let totalProjectTime = 0;
+      let enteredTimeEntries = 0;
 
-//     const weeklyCombinedData = TimebankUtilities.combineWeeklyData(personTotalTimes, slackUsers);
+      if (workWeek) {
+        for (const day of workWeek) {
+          console.log("day", day.enteredHours);
```
naStipe: this is uncecessary

src/meta-assistant/generic/time-utils.ts
```diff
@@ -53,10 +52,30 @@ namespace TimeUtilities {
...
+
+  /**
+   * Handle formatting multiple time variables for weekly data
+   *
+   * @param user data from severa
+   * 
+   * @returns human friendly time formats
+   */
+  export const handleTimeFormattingWeekly = (user: TotalTimeWeekly): DisplayValuesWeekly => {
+    const { totalEnteredHours, totalExpectedHours, projectTime   } = user;
```
naStipe, line 69: not sure if we should name some time variables as "user"

src/types/meta-assistant/index.ts
```diff
@@ -28,19 +27,26 @@ export interface DailyCombinedData {
...
 export interface WeeklyCombinedData {
-  selectedWeek: PersonTotalTime;
-  name: string;
+  userId: string;
   firstName: string;
-  slackId?: string;
-  personId: number;
-  expected: number;
+  totalExpectedHours: number;
+  // totalExpectedHours: number;
```
naStipe: can this be deleted if unused?

naStipe (approved):

src/meta-assistant/generic/time-utils.ts
Hanhz2707: it's not a time variables. The function will be passsed an user to its param. The user's data will consist some "time" value. All the props from the params are from the user data. TotalTime here i guess is just meaning it's containing all kind of time that we need such as totalLoggedTime or expectedHours,.... Hopefully make it clearer

rimz178 (approved):

StudVladimir (approved):

MuhammadKashif064 (approved):

raorizwan1092 (approved):

src/meta-assistant/slack/slack-utils.ts
```diff
@@ -242,35 +235,25 @@ Have a great week!
...
     for (const userData of weeklyCombinedData) {
-      const { slackId, personId, expected } = userData;
-      const vacationTime = TimeUtilities.checkIfVacationCaseExists(personId, timeRegistrations, nonProjectTimes, weekStartDate, weekEndDate);
-      const isAway = TimeUtilities.checkIfUserShouldRecieveMessage(timeRegistrations, personId, expected, today.toISODate(), nonProjectTimes);
-      const firstDayBack = TimeUtilities.checkIfUserShouldRecieveMessage(timeRegistrations, personId, expected, yesterday.toISODate(), nonProjectTimes);
-      const message = constructWeeklySummaryMessage(userData, weekStartDate.toISODate(), weekEndDate.toISODate(), vacationTime);
+      const { userId } = userData;
+      const message = constructWeeklySummaryMessage(userData, weekStartDate.toISODate(), weekEndDate.toISODate());
 
-      if (!isAway && !firstDayBack) {
```
Jdallos, line 261: We might want to restore this, as we don't want to send the mesage to users when they are absent (or on their first day back- as they won't have worked on the previous day and so there is nothing to show).
DIg into the code and the checkIfUserShouldRecieveMessage function to make sure this should work. We can test it together by updating the demo data next week

src/services/severa-api-service.ts
```diff
@@ -156,6 +158,38 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+     * @param severaUserId Severa user id
+     * @returns Workdays of a user
+     */
+    getWorkWeek: async (severaUserId: string) => {
+
+      const today = DateTime.now().toISODate();
+      const weekAgo = DateTime.now().minus({ days: 7 }).toISODate();
+      const isProduction = process.env.NODE_ENV === "production";
+      const endDate = isProduction ? today : "2025-01-20";
+      const startDate = isProduction ? weekAgo : "2025-01-13";
```
Jdallos: We will need to update this in order to be able to test with demo data for the current weeek. I suggest adding an env locally that provides the date (or doesn't). We could then replace this isProduction check with a Demo_Data_Date env, if it's not present we know we are in production, but it also allows us to control the data for demo data we check rather than have to hardcode it like this.

src/services/severa-api-service.ts
```diff
@@ -213,6 +247,35 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+        throw new Error(
+          `Failed to fetch work hours: ${response.status} - ${response.statusText}`,
+        );
+      }
+      return response.json();
+    },
+
+    /**
+     * Gets previous workweeks Workhours from Severa
+     */
```
Jdallos: @param missing from Docs

src/services/severa-api-service.ts
```diff
@@ -213,6 +247,35 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+    getPreviousWeekHours: async (severaUserId: string) => {
+      const weekAgo = DateTime.now().minus({ days: 7 }).toISODate();
+      const today = DateTime.now().toISODate();
+      const isProduction = process.env.NODE_ENV === "production";
+      const startDate = isProduction ? weekAgo : "2025-01-13";
+      const endDate = isProduction ? today : "2025-01-20";
```
Jdallos: Again this can be replaced in the same way as described previously- maybe needs to be another seperate env?
Once this has been done and tested locally we need to add the envs to hcv before merging

src/types/severa/workDays/severaResponseWorkDays.ts
```diff
@@ -5,6 +5,8 @@ interface SeveraResponseWorkDays {
   userGuid: string;
   expectedHours: number;
   enteredHours: number;
+  enteredTimeEntries: number;
+
```
Jdallos:
```suggestion
```

Jdallos (changes requested):

src/services/severa-api-service.ts
```diff
@@ -213,6 +247,38 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+
+    /**
+     * Gets previous workweeks Workhours from Severa
+     *
+     * @param severaUserId Severa user id
+     * @returns Previous workweek workhours for the user
+     */
+    getPreviousWeekHours: async (severaUserId: string) => {
+      const demoDataDate = process.env.DEMO_DATA_DATE;
+      const today = demoDataDate ? DateTime.fromISO(demoDataDate).toISODate() : DateTime.now().toISODate();
```
Jdallos, line 273: This is good, lets do the same for the getPreviousWorkHours function to replace the isProduction env.
I think it should be fine to use the same env, although if that env isn't set as 2024-11-26 then I guess we can expect the data to change. But that's not an issue as we can reconfigure the demo data today together.

Jdallos (approved):

Jdallos (approved):

PR #168, opened 2025-04-07
Thesis Internal Wiki for documentation management 

Kseniia-Koshkina:


src/functions/wiki-documentation/find-article/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from '@libs/handler-resolver';
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: 'get',
+        path: '/articles/{path+}',
```
anttileppa: Find endpoints should not accept path. Instead they should use ID, so instead of doing this you should have 

list endpoint with query parameter path:

    /articles?path=something

and find endpoint with:

    /articles/{articleId}

serverless.ts
```diff
@@ -206,9 +225,20 @@ const serverlessConfiguration: AWS = {
...
+    updateArticleHandler,
+    deleteArticleHndler,
+    readArticleHandler,
+    uploadFileHandler
   },
   package: { individually: true },
   custom: {
+    s3BucketName: {
+      dev: "home-staging",
+      production: "home"
```
anttileppa: These shoul.d be environment variables instead of hard coded values

src/functions/wiki-documentation/upload-file/handler.ts
```diff
@@ -0,0 +1,42 @@
...
+      statusCode: 400,
+      body: JSON.stringify({
+        code: 400,
+        message: "Invalid request body.",
+      })
+    };
+  }
+
+  try {
+    const presignedUrl = await generatePreSignedUrl(path, "put", contentType);
```
anttileppa: Why the presigned request is created in the backend? Usually these are done in the UI and we are using seperate lambda for uploading the actual files

anttileppa (changes requested):

Jdallos (approved):

PR #171, opened 2025-04-17
Vacation management using keyloak-services routes handling

MuhammadKashif064:


balashova12 (approved):

AYZBTR (approved):

serverless.ts
```diff
@@ -1,7 +1,7 @@
 import type { AWS } from "@serverless/typescript";
 import * as dotenv from "dotenv";
-dotenv.config({ path: __dirname + "/.env" });
-import { env } from "process";
+dotenv.config({ path: `${__dirname}/.env` });
+import { env } from "node:process";
```
Jdallos: Are you sure these changes were necessary as they seemed to be working well up until now?
If yes please explain why, otherwise revert to how they were originally.

serverless.ts
```diff
@@ -25,6 +25,7 @@ import listUsersHandler from "@/functions/keycloak/list-users";
 import findUserHandler from "@/functions/keycloak/find-user";
 import updateUserAttributeHandler from "@/functions/keycloak/update-user-attributes";
 import removeUserAttributeHanndler from "src/functions/keycloak/remove-user-attribute";
+import updatevacationhandler from "src/functions/keycloak/update-user-vacation";
```
Jdallos: This should be named using camel case like the other handleres e.g. updateVacationHandler

serverless.ts
```diff
@@ -183,6 +203,7 @@ const serverlessConfiguration: AWS = {
     findUserHandler,
     updateUserAttributeHandler,
     removeUserAttributeHanndler,
+    updatevacationhandler,
```
Jdallos: camelCase

src/database/services/keycloak-api-service.ts
```diff
@@ -158,70 +166,107 @@ export const CreateKeycloakApiService = (): KeycloakApiService => {
...
+          throw new Error(
+            `Failed to update user attributes: ${updateResponse.status} - ${updateResponse.statusText}. Details: ${errorText}`
+          );
+        }
+      } catch (error) {
+        throw new Error(
+          error instanceof Error
+            ? error.message
+            : "An unknown error occurred while removing the user attribute."
+        );
```
Jdallos, line 234: These are showing up as changes but have you actually made changes to this function or is it just linter/ indentation related changes? At least as far as I can tell you don't seem to be using this code in this vacation management feature.

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,110 @@
+import { middyfy } from "src/libs/lambda";
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import type { APIGatewayProxyHandlerV2 } from "aws-lambda";
+
+const keycloakApiService = CreateKeycloakApiService();
+
+const corsHeaders = {
+  "Content-Type": "application/json",
+  "Access-Control-Allow-Origin": "*",
+};
```
Jdallos: I don't think this is necessary, it's not used for any of the other lambdas and I don't think we need to define these for the lambdas- they are actually defined as part of the API gateways COrs configuration in the aws.
So you should be able to remove these corsHeaders and all references to them, no need to pass headers as part of the responses.

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const updatevacationhandler: APIGatewayProxyHandlerV2 = async (event) => {
```
Jdallos: Missing TSDocs

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,110 @@
...
+      statusCode: 400,
+      headers: corsHeaders,
+      body: JSON.stringify({ message: "User ID is required" }),
+    };
+  }
+
+  try {
+    const requestBody =
+      typeof event.body === "string" ? JSON.parse(event.body) : event.body;
+    const { vacationDays } = requestBody;
```
Jdallos: This is currently types as any, as we are using typescript we should avoid uses of any where possible. You have a few options, you could more comprehensively type the event the handler is receiving and type the contents of the body and then the vacationDays. Or you could just type/ cast the vacationDays as you desctrucure them here.

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,110 @@
...
+    const unspentVacationDaysByYear =
+      userAttributes.unspentVacationDaysByYear || [];
+
+    // Store updated vacation days in a separate object
+    const updatedVacationDays: Record<
+      string,
+      { total: number; remaining: number }
+    > = {};
+
+    Object.entries(vacationDays).forEach(([year, data]: [string, any]) => {
```
Jdallos: Again we need the type for vacationDays here and there should also be some validation done to ensure it is of the expected structure.

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (!vacationDays) {
+      return {
+        statusCode: 400,
+        headers: corsHeaders,
+        body: JSON.stringify({ message: "Vacation days are required" }),
+      };
+    }
+
+    const userAttributes = await keycloakApiService.getUserAttributes(userId);
+    const updatedAttributes = { ...userAttributes };
```
Jdallos: Is this actually necessary to clone this like this, can't you just use userAttributes as is?

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const vacationDaysByYear = userAttributes.vacationDaysByYear || [];
```
Jdallos, line 43: It's also maybe a bit odd you clone userAttributes as updatedAttributes but then use userAttributes again here. Perhaps you can just have userAttributes without needing to clone it into updatedAttributes

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    // Store updated vacation days in a separate object
+    const updatedVacationDays: Record<
+      string,
+      { total: number; remaining: number }
+    > = {};
+
+    Object.entries(vacationDays).forEach(([year, data]: [string, any]) => {
+      const yearInt = Number.parseInt(year, 10);
+      if (!Number.isNaN(yearInt) && yearInt <= new Date().getFullYear()) {
```
Jdallos: I'd suggest creating a named variable from:
new Date().getFullYear()

calling it currentyear.

Then you have yearInt <= currentYear which is clearer

src/functions/keycloak/update-user-vacation/handler.ts
Jdallos: Can we type the data to avoid the use of any?

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,110 @@
...
+          total: data.total,
+          remaining: data.remaining,
+        };
+
+        updatedAttributes[`vacation_${year}`] = [String(data.total)];
+        updatedAttributes[`vacation_${year}_remaining`] = [
+          String(data.remaining),
+        ];
+      }
+    });
```
Jdallos: This function is a little dense, perhaps it could be refactored to improve readability, perhaps extract the updateEntry into a separate function, also extract the padstart and fullYear, here is an ai suggestion (you should test):

`const currentYear = new Date().getFullYear();

const pad = (num: number) => String(num).padStart(3, "0");

const updateEntry = (arr: string[], year: string, value: string) => {
  const prefix = `${year}:`;
  const index = arr.findIndex((v) => v.startsWith(prefix));
  if (index !== -1) arr[index] = value;
  else arr.push(value);
};

Object.entries(vacationDays).forEach(([year, { total, remaining }]) => {
  const yearInt = parseInt(year, 10);
  if (isNaN(yearInt) || yearInt > currentYear) return;

  const formattedTotal = `${year}:${pad(total)}`;
  const formattedRemaining = `${year}:${pad(remaining)}`;

  updateEntry(vacationDaysByYear, year, formattedTotal);
  updateEntry(unspentVacationDaysByYear, year, formattedRemaining);

  updatedVacationDays[year] = { total, remaining };

  updatedAttributes[`vacation_${year}`] = [String(total)];
  updatedAttributes[`vacation_${year}_remaining`] = [String(remaining)];
});
`

src/tests/serverless/mockData.json
```diff
@@ -2,7 +2,7 @@
   "headers": {
     "Content-Type": "application/json"
   },
-  "body": "{\"email\": \"*insert the email of the user whose attributes are to be updated.*\"}",
+  "body": "{\"email\": \"test-user@example.fi\"}",
```
Jdallos: Lets leave the mock data generic, please revert to have the instructions here rather than the test user email

Jdallos (changes requested):

serverless.ts
MuhammadKashif064: Reverted back to the previous change

src/database/services/keycloak-api-service.ts
MuhammadKashif064: yes, i did not use this code but i just a auto format on save

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -4,101 +4,113 @@ import type { APIGatewayProxyHandlerV2 } from "aws-lambda";
...
-const updatevacationhandler: APIGatewayProxyHandlerV2 = async (event) => {
+
+interface VacationDayEntry {
+  total: number;
+  remaining: number;
+}
+
+interface VacationDays {
+  [year: string]: VacationDayEntry;
+}
```
Jdallos: These could be moved to and exported from types/vacations/index.ts

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -4,101 +4,113 @@ import type { APIGatewayProxyHandlerV2 } from "aws-lambda";
...
       typeof event.body === "string" ? JSON.parse(event.body) : event.body;
-    const { vacationDays } = requestBody;
 
-    if (!vacationDays) {
+    const { vacationDays } = requestBody as { vacationDays: VacationDays };
+
+    if (
+      !vacationDays ||
+      typeof vacationDays !== "object" ||
+      Array.isArray(vacationDays)
```
Jdallos, line 33: Are these 2 checks really necessary?

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -4,101 +4,113 @@ import type { APIGatewayProxyHandlerV2 } from "aws-lambda";
...
-          String(data.remaining),
-        ];
+      if (
+        Number.isNaN(yearInt) ||
+        yearInt > currentYear ||
+        typeof data.total !== "number" ||
+        typeof data.remaining !== "number"
+      ) {
+        return;
       }
```
Jdallos: Are all of these checks really necessary?
Should there be some kind of log or error thrown if this happens?

Jdallos (approved):

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,122 @@
+import { middyfy } from "src/libs/lambda";
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import type { APIGatewayProxyHandlerV2 } from "aws-lambda";
+
+const keycloakApiService = CreateKeycloakApiService();
+
+
```
Jdallos: Remove unnecessary line breaks

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,122 @@
...
+      userAttributes.unspentVacationDaysByYear || [];
+
+    const currentYear = new Date().getFullYear();
+
+    const updatedVacationDays: Record<string, VacationDayEntry> = {};
+
+    const formatValue = (year: string, value: number) =>
+      `${year}:${String(value).padStart(3, "0")}`;
+
+    const updateEntry = (arr: string[], value: string) => {
```
Jdallos, line 76: Docs/ comments to clarify this function would be helpful

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,122 @@
...
+      const index = arr.findIndex((v) => v.startsWith(`${value.split(":")[0]}:`));
+      if (index !== -1) arr[index] = value;
+      else arr.push(value);
+    };
+
+    Object.entries(vacationDays).forEach(([year, data]) => {
```
Jdallos, line 82: Docs/ comments to clarify here would be helpful

src/functions/keycloak/update-user-vacation/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "put",
+        path: "/users/{userId}/vacation",
```
Jdallos: I'm not certain about this endpoint name. Might vacationDays or vacationDaysRecord be better/ clearer?

Jdallos (approved):

src/functions/keycloak/update-user-vacation/handler.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    await keycloakApiService.updateUserAttributes(userId, userAttributes);
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify({
+        id: userId,
+        vacationDays: updatedVacationDays,
+      }),
+    };
```
Jdallos, line 122: This is a fine response, but it does not match the response indicated for this endpoint in the spec, I'd suggest sticking with this and updating the spec

src/functions/keycloak/update-user-vacation/index.ts
Jdallos: Changing the name will also require spec updates

Jdallos (approved):

PR #175, opened 2025-05-22
Update getFlextimeHandler Lambda to prevent access unless user has isSeveraOptIn keyword

balashova12:


AYZBTR (approved):

MuhammadKashif064 (approved):

raorizwan1092 (approved):

StudVladimir (approved):

src/services/severa-api-service.ts
```diff
@@ -224,7 +225,32 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+          "Content-Type": "application/json"
+        }
+      });
+
+      if (!response.ok) {
+        throw new Error(`Failed to fetch Severa user: ${response.status} - ${response.statusText}`);
+      }
+
+      return response.json();
+    },
```
Jdallos: Is it not possible to actually use the keyword here so that we don't even receive data in the lambda unless they are opted in.

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -20,18 +24,48 @@ export const getFlextimeHandler: APIGatewayProxyHandler = async (event) => {
...
+    const isOptedIn = severaUser.keywords?.some(
+      (keyword) => keyword.value === "isSeveraOptIn"
+    );
+
+    if (!isOptedIn) {
+      return {
+        statusCode: 403,
+        body: JSON.stringify({ message: "User has not opted in to flextime tracking." }),
+      };
+    }
```
Jdallos: I'm pretty sure we can pass the keyword to the severa request to filter by keyword straight away. This is the best way so that we don't even get any user data (in the lambda) unless they have opted in.

This will need changes to the severa-api-service most likely

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -20,18 +24,48 @@ export const getFlextimeHandler: APIGatewayProxyHandler = async (event) => {
...
+
+    const keycloakEmail = event.requestContext?.authorizer?.claims?.email || "test-user@example.fi";
```
Jdallos: Does this actually work for you in staging? Did you try invoking this locally using the credentials you have when logged into the UI? Won't it still use your keycloak email, which will not match the staging testuser email in severa?

We need to think of some way of identifying if the UI is in staging and if so it should always use the test-user email. If not it should always use the authorizer email as you have done here.

My suggestion is that we should make sure severaTestUser email env is optional and if it is present it is used here and if it is not it uses the actual email.
Then in the pipelines I will add the test user email for staging lambdas and not for the production lambda. Does that sound reasonable?

src/services/severa-api-service.ts
Jdallos: Check the severa API docs, I'm quite sure this should be possible

Jdallos (changes requested):

src/services/severa-api-service.ts
```diff
@@ -224,7 +225,33 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+    
+    /**
+   * Gets a specific user from Severa by their user ID.
+   *
+   * @param severaUserId - Severa user ID
+   * @returns Severa user object, including keywords
+   */
+    getUser: async (severaUserId: string) => {
+    const optInKeywordId = "8e7b363e-aa8c-34b1-478f-0a9633848fde";
```
Jdallos: This is the only bit I'm unsure about.
Hardcoding this optInKeywordId seems reasonable for now assuming the id does not change. I think it's the same id/ keyword used for users **however** when we switch from the severa DEMO client id to the "production severa client" I assume we will create a new keyword which will have it's own uuid, meaning this won't work.

So it's a bit of a pain but for flexibility I would make some getKeywordId method in the severa api service.
It should take the text of the keyword (for now we will only user severaOptIn, or whatever it is, but it's good to build in flexibility). The function should then use the users/userId/keywords endpoint from severa to list the keywords, match the provided string and return the valid id. This means it will work across environments.

Related to this we should investigate how is the keyword being created, is it being created as part of the opt in flow e.g. if severaOptIn does not exist, it is created. If not it should be. This could be a separate task/ branch  and could also be passed to/ worked on with another trainee if you want.

Let me know how you want to approach this and if anything needs clarifying

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -20,18 +24,38 @@ export const getFlextimeHandler: APIGatewayProxyHandler = async (event) => {
...
+    const severaUser = await api.getUser(severaUserId);
+    
+    if (!severaUser) {
+      return {
+        statusCode: 404,
+        body: JSON.stringify({ message: "Severa user not found." }),
+      };
+    }
+
+    const testUserEmail = process.env.SEVERA_TEST_USER_EMAIL;
```
Jdallos: Will this cause the app to crash in production if there is no SEVERA_TEST_USER_EMAIL.
We need to make sure this is an optional env so you should probably look into the the serverless framework docs to make sure we use this correctly

Jdallos (changes requested):

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -20,18 +22,40 @@ export const getFlextimeHandler: APIGatewayProxyHandler = async (event) => {
...
     }
 
-    const flexTimeBySeveraId = await api.getFlextimeBySeveraUserId(severaUserId);
+    try {
+      await api.getKeywordIdForUser(severaUserId, "isSeveraOptIn");
+    } catch {
+      return {
+        statusCode: 403,
+        body: JSON.stringify({ message: "User has not opted in." }),
+      };
```
Jdallos, line 30: Maybe its a bit confusing that we go to the catch here, to me it would be clearer if we checked the response, if it was empty (or false) then we would return the 403 and message.
But you have set the request to throw an Error if user is not found and it works fine. So keep this or change, up to you.

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -20,18 +22,40 @@ export const getFlextimeHandler: APIGatewayProxyHandler = async (event) => {
...
+    }
+
+    const severaUser = await api.getUser(severaUserId);
+
+    const testUserEmail = Config.get().testUser.email;
+    const keycloakEmail = testUserEmail || event.requestContext?.authorizer?.claims?.email;
```
Jdallos: What about in production where we don't want to use the testUser email.
I'm concerned it would crash if the env is missing- so we need to make sure it doesn't (test locally what happens).

Jdallos (approved):

src/functions/severa/get-flextime-by-user/handler.ts
```diff
@@ -1,14 +1,16 @@
...
 /**
- * Handler for getting flextime by user from Severa REST API.
- * 
- * @param event - API Gateway event containing the user Id.
+ * Lambda handler for getting flextime by severaUserId,
+ * but only if the user has opted in (has "isSeveraOptIn" keyword).
  */
 export const getFlextimeHandler: APIGatewayProxyHandler = async (event) => {
   const severaUserId = event.pathParameters?.severaUserId;
+  const eventDate = DateTime.now().toISODate();
```
Jdallos: If you look at the severa request method this is being passed to it is no longer being used, so remove it from the request types and then it can be removed from here.

Jdallos (approved):

Jdallos (commented):
Approved

PR #181, opened 2025-06-23
Feature-179-Sprint-view-optIn-changes

StudVladimir:
### Function: `getWorkHoursHandler` (`src/functions/severa/get-filtered-workhours/handler.ts`)
**Purpose:**  
Returns Severa users' work hours, with filtering by project, phase, user, and includes only opted-in users.

---

#### Filtering Only Opted-In Users
- Retrieves a list of users who have opted in for data processing.
- All requests to the Severa API are made **only** for these users.

#### Flexible Parameter Filtering
- Supports the following filters: `severaProjectId`, `severaPhaseId`, `severaUserId`, `startDate`, `endDate`.
- If `severaProjectId` is specified (and `severaUserId` is not), a **separate API request** is made for each opted-in user, filtered by project.
- If `severaUserId` is specified, the API request is made **only for that user**.
- If neither project nor user is specified, a general endpoint is used (defined in the code, but may be commented out).

#### Explicit Project Filtering
- After retrieving all work hours for opted-in users, the results are **further filtered** to include only records where `project.guid` strictly matches the provided `severaProjectId`.
- This ensures no extraneous records are returned, even if Severa API or mock data includes more than necessary.

#### Phase Filtering
- If `severaPhaseId` is provided, work hours are further filtered by the corresponding project phase.

**How to test getWorkHoursHandler?**

To run test on local backend, you can run - `npx sls invoke local -f getWorkHoursHandler --stage dev --path src/tests/serverless/mockData.json`

mockData:

```
"queryStringParameters": {
    "severaProjectId": "PROJECT_ID"
  }
```

Possible project IDs:
```
f7d6b20f-20a7-8abc-b9af-527537a6f197
7f981300-a213-bdd3-5a2c-2f05441ce29a
334a66c6-cb35-fb03-b209-18626a31047e
```
---
### Function: `getResourceAllocationHandler` (`handler.ts`)
**Purpose:**  
Returns Severa users' resource allocations, including only opted-in users. Supports filtering by user.

---

#### Filtering Only Opted-In Users
- Retrieves a list of users who have explicitly opted in.
- All requests to the Severa API are made **only for these users**.

#### Flexible User Filtering
- Supports the `severaUserId` filter (via query string).
- If `severaUserId` is provided, the Severa API request is made **only for that user**.
- If `severaUserId` is not provided, a **separate request** is made for **each** opted-in user, and the results are combined.

#### Batch Requests for All Opted-In Users
- To retrieve resource allocations for **all** opted-in users, the function makes **parallel API requests** to Severa for each user.
- All results are merged into a single array before being returned to the client.

#### Consistent Data Mapping
- After fetching data from the Severa API, all resource allocations are mapped into a unified internal format: `ResourceAllocationModel`.

**How to test getResourceAllocationHandler?**

To run test on local backend, you can run - `npx sls invoke local -f getResourceAllocationHandler --stage dev --path src/tests/serverless/mockData.json`

For testing through **POSTMAN**, run backend localy and use this url - `http://localhost:3000/dev/severa/resourceAllocations?severaUserId`
`http://localhost:3000/dev/severa/resourceAllocations?severaUserId=USER_ID`

### DIFF
--- a/src/functions/severa/get-filtered-workhours/handler.ts
+++ b/src/functions/severa/get-filtered-workhours/handler.ts
@@ -26,10 +26,12 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
     } else {
       if (severaUserId) {
         if (!optInUserGuids.has(severaUserId)) {
-          allWorkHours = [];
-        } else {
-          allWorkHours = await api.getWorkHoursForUser(severaUserId, startDate, endDate);
+          return {
+            statusCode: 404,
+            body: JSON.stringify({ error: "User not found or not opted-in" }),
+          };
         }
+        allWorkHours = await api.getWorkHoursForUser(severaUserId, startDate, endDate);
       } else {
         throw new Error("severaUserId is required when severaProjectId is not specified");
       }
--- a/src/functions/severa/get-filtered-workhours/handler.ts
+++ b/src/functions/severa/get-filtered-workhours/handler.ts
@@ -19,6 +19,7 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
     const optInUsers = await api.getOptInUsers();
     const optInUserGuids = new Set(optInUsers.map((u) => u.guid));
 
+    let errors: Array<{ userGuid: string, error: any }> = [];
     // If severaProjectId is specified and severaUserId is not, make requests for each opted-in user
     let allWorkHours: SeveraResponseWorkHours[] = [];
     if (severaProjectId) {
@@ -33,7 +34,7 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
             const userWorkHours = await api.getWorkHours(url);
             return userWorkHours;
           } catch (e) {
-            // If there is an error for a single user, just skip
+            errors.push({ userGuid: user.guid, error: e });
             return [];
           }
         })
--- a/src/functions/severa/get-filtered-workhours/handler.ts
+++ b/src/functions/severa/get-filtered-workhours/handler.ts
@@ -30,12 +30,23 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
      *
      * @returns {string} - The constructed custom URL for fetching work hours.
      */
-    const buildWorkHoursUrl = (severaProjectId?: string, severaUserId?: string, startDate?: string, endDate?: string) => {
+    const buildWorkHoursUrl = async (severaProjectId?: string, severaUserId?: string, startDate?: string, endDate?: string) => {
       let endpointPath: string;
+      try {
+      await api.getKeywordIdForUser(severaUserId, "isSeveraOptIn");
+    } catch {
+      return {
+        statusCode: 403,
+        body: JSON.stringify({ message: "User has not opted in." }),
+      };
+    }
 
       if (severaProjectId) {
         endpointPath = `projects/${severaProjectId}/workhours`;
       } else if (severaUserId) {
+      if (severaUserId && !optInUserGuids.has(severaUserId)) {
+        throw new Error("User has not opted in.");
+      }
         endpointPath = `users/${severaUserId}/workhours`;
       } else {
         endpointPath = "workhours";
@@ -49,7 +60,11 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
       return customUrl;
     };
 
-    const url = buildWorkHoursUrl(severaProjectId, severaUserId, startDate, endDate);
+    const url = await buildWorkHoursUrl(severaProjectId, severaUserId, startDate, endDate);
+    if (!(url instanceof URL)) {
+      // url is an error response object, return it directly
+      return url;
+    }
     const response = await api.getWorkHours(url);
 
     // Filtering by opted in users only
--- a/src/functions/severa/get-filtered-workhours/handler.ts
+++ b/src/functions/severa/get-filtered-workhours/handler.ts
@@ -11,7 +11,7 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
  * @param event - API Gateway event containing the userId, projectId, and phaseId.
  */
 export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
-  const { startTime: startTime, endTime: endTime, severaProjectId, severaPhaseId, severaUserId } = event.queryStringParameters || {};
+  const { startDate: startDate, endDate: endDate, severaProjectId, severaPhaseId, severaUserId } = event.queryStringParameters || {};
 
   try {
     const api = CreateSeveraApiService();
@@ -30,8 +30,8 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
             const url = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/users/${user.guid}/workhours`);
             // This endpoint/query parameter is working but it is not documented proper in Severa API spec
             url.searchParams.append("projectGuid", severaProjectId);
-            if (startTime) url.searchParams.append("startDate", startTime);
-            if (endTime) url.searchParams.append("endDate", endTime);
+            if (startDate) url.searchParams.append("startDate", startDate);
+            if (endDate) url.searchParams.append("endDate", endDate);
             const userWorkHours = await api.getWorkHours(url);
             return userWorkHours;
           } catch (e) {
@@ -55,8 +55,8 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
       } else {
         throw new Error("severaUserId is required when severaProjectId is not specified");
       }
-      if (startTime) url.searchParams.append("startTime", startTime);
-      if (endTime) url.searchParams.append("endTime", endTime);
+      if (startDate) url.searchParams.append("startDate", startDate);
+      if (endDate) url.searchParams.append("endDate", endDate);
 
       allWorkHours = await api.getWorkHours(url);
     }
--- a/src/functions/severa/get-filtered-workhours/handler.ts
+++ b/src/functions/severa/get-filtered-workhours/handler.ts
@@ -11,7 +11,7 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
  * @param event - API Gateway event containing the userId, projectId, and phaseId.
  */
 export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
-  const { startDate, endDate, severaProjectId, severaPhaseId, severaUserId } = event.queryStringParameters || {};
+  const { startTime: startTime, endTime: endTime, severaProjectId, severaPhaseId, severaUserId } = event.queryStringParameters || {};
 
   try {
     const api = CreateSeveraApiService();
@@ -30,8 +30,8 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
             const url = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/users/${user.guid}/workhours`);
             // This endpoint/query parameter is working but it is not documented proper in Severa API spec
             url.searchParams.append("projectGuid", severaProjectId);
-            if (startDate) url.searchParams.append("startDate", startDate);
-            if (endDate) url.searchParams.append("endDate", endDate);
+            if (startTime) url.searchParams.append("startDate", startTime);
+            if (endTime) url.searchParams.append("endDate", endTime);
             const userWorkHours = await api.getWorkHours(url);
             return userWorkHours;
           } catch (e) {
@@ -53,11 +53,10 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
       if (severaUserId) {
         url = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/users/${severaUserId}/workhours`);
       } else {
-      // Absolute endpoint, fetching all work hours
-      url = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/workhours`);
+        throw new Error("severaUserId is required when severaProjectId is not specified");
       }
-      if (startDate) url.searchParams.append("startDate", startDate);
-      if (endDate) url.searchParams.append("endDate", endDate);
+      if (startTime) url.searchParams.append("startTime", startTime);
+      if (endTime) url.searchParams.append("endTime", endTime);
 
       allWorkHours = await api.getWorkHours(url);
     }
--- a/src/functions/severa/get-filtered-workhours/handler.ts
+++ b/src/functions/severa/get-filtered-workhours/handler.ts
@@ -16,6 +16,10 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
   try {
     const api = CreateSeveraApiService();
 
+    // Getting the list of opted in users
+    const optInUsers = await api.getOptInUsers();
+    const optInUserGuids = new Set(optInUsers.map((u) => u.guid));
+
     /**
      * Construct a custom url for fetching workHours with required queryParams.
      *
@@ -48,11 +52,15 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
     const url = buildWorkHoursUrl(severaProjectId, severaUserId, startDate, endDate);
     const response = await api.getWorkHours(url);
 
+    // Filtering by opted in users only
     const filteredWorkHours = response.filter((workHours: SeveraResponseWorkHours) => {
-    /**
-     * Note: If severaProjectId and severaUserId are both true, the work hours data
-     * will be filtered by project in the Severa API call due to workHours custom url.
-     */
+      if (!workHours.user?.guid || !optInUserGuids.has(workHours.user.guid)) {
+        return false;
+      }
+      /**
+       * Note: If severaProjectId and severaUserId are both true, the work hours data
+       * will be filtered by project in the Severa API call due to workHours custom url.
+       */
       if (severaProjectId && severaUserId) {
         const filteredWorkHoursUserProject = FilterUtilities.filterByUserSevera(workHours.user?.guid, severaUserId);
         if (!filteredWorkHoursUserProject) return false ;
--- a/src/functions/severa/get-filtered-workhours/handler.ts
+++ b/src/functions/severa/get-filtered-workhours/handler.ts
@@ -15,75 +15,60 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
 
   try {
     const api = CreateSeveraApiService();
-
     // Getting the list of opted in users
     const optInUsers = await api.getOptInUsers();
     const optInUserGuids = new Set(optInUsers.map((u) => u.guid));
 
-    /**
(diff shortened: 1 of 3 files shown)


src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -15,52 +15,60 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
...
+    } else {
+      // Standard logic for other cases (by userId or without projectId)
+      // Manually build the URL since buildWorkHoursUrl does not exist
+      let url: URL;
+      if (severaUserId) {
+        url = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/users/${severaUserId}/workhours`);
       } else {
-        endpointPath = "workhours";
+      // Absolute endpoint, fetching all work hours
+      url = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/workhours`);
```
StudVladimir: This is Absolut endpoint which return all possible values without any filtering. If you are testing code using parametrs, you should comment this line.

For future this line would be deleted due to the possibility of requesting data from users who have not passed the OptedIn check. Safe in a test environment, but be sure to remove before production.

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -15,52 +15,60 @@ export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
...
       }
+      if (severaProjectId) url.searchParams.append("projectGuid", severaProjectId);
+      if (startDate) url.searchParams.append("startDate", startDate);
+      if (endDate) url.searchParams.append("endDate", endDate);
```
StudVladimir: Now when testing I do not recommend adding parameters start date and end date because they do not exist in the current configuration

StudVladimir:
**How to test getWorkHoursHandler?**

To run test on local backend, you can run - `npx sls invoke local -f getWorkHoursHandler --stage dev --path src/tests/serverless/mockData.json`

mockData:

```
"queryStringParameters": {
    "severaProjectId": "PROJECT_ID"
  }
```

Possible project IDs:
```
f7d6b20f-20a7-8abc-b9af-527537a6f197
7f981300-a213-bdd3-5a2c-2f05441ce29a
334a66c6-cb35-fb03-b209-18626a31047e
```

MuhammadKashif064 (approved):

StudVladimir:
**How to test getResourceAllocationHandler?**

To run test on local backend, you can run - `npx sls invoke local -f getResourceAllocationHandler --stage dev --path src/tests/serverless/mockData.json`

For testing through **POSTMAN**, run backend localy and use this url - `http://localhost:3000/dev/severa/resourceAllocations?severaUserId`
`http://localhost:3000/dev/severa/resourceAllocations?severaUserId=USER_ID`

StudVladimir:
### Function: `getResourceAllocationHandler` (`handler.ts`)
**Purpose:**  
Returns Severa users' resource allocations, including only opted-in users. Supports filtering by user.

---

### Key Features

#### Filtering Only Opted-In Users
- Retrieves a list of users who have explicitly opted in.
- All requests to the Severa API are made **only for these users**.

#### Flexible User Filtering
- Supports the `severaUserId` filter (via query string).
- If `severaUserId` is provided, the Severa API request is made **only for that user**.
- If `severaUserId` is not provided, a **separate request** is made for **each** opted-in user, and the results are combined.

#### Batch Requests for All Opted-In Users
- To retrieve resource allocations for **all** opted-in users, the function makes **parallel API requests** to Severa for each user.
- All results are merged into a single array before being returned to the client.

#### Consistent Data Mapping
- After fetching data from the Severa API, all resource allocations are mapped into a unified internal format: `ResourceAllocationModel`.

src/functions/severa/get-filtered-workhours/handler.ts
Jdallos: Why not remove this already as we have discussed that this endpoint is not suitable for our needs as we can't ensure it is only fetching opted in Users?

src/functions/severa/get-filtered-workhours/handler.ts
Jdallos: If there parameters don't exist/ don't work then they should be removed. However I though we looked at the spec together and found that they did exist (at least for some of the severa endpoints you are requesting) but they were named slightly different- if this is the case then the naming should be updated to use the correct name.

StudVladimir: I was wrong, the parameters were correct - startDate/endDate has to be used. They are using as parameters for filtering output, but they are not shown in output, that's why I was confused.

balashova12 (approved):

gun-girl (approved):

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
  * @param event - API Gateway event containing the userId, projectId, and phaseId.
  */
 export const getWorkHoursHandler: APIGatewayProxyHandler = async (event) => {
-  const { startDate, endDate, severaProjectId, severaPhaseId, severaUserId } = event.queryStringParameters || {};
+  const { startDate: startDate, endDate: endDate, severaProjectId, severaPhaseId, severaUserId } = event.queryStringParameters || {};
```
Jdallos: This renaming is redundant for startDate and endDate

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
 
   try {
     const api = CreateSeveraApiService();
+    // Getting the list of opted in users
```
Jdallos:
```suggestion
```
Unnecessary comment

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
+    const optInUsers = await api.getOptInUsers();
+    const optInUserGuids = new Set(optInUsers.map((u) => u.guid));
```
Jdallos, line 19: No harm in using a set here, although I wouldn't expect it to be necessary.

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
-
-      if (severaProjectId) {
-        endpointPath = `projects/${severaProjectId}/workhours`;
-      } else if (severaUserId) {
-        endpointPath = `users/${severaUserId}/workhours`;
+    let errors: Array<{ userGuid: string, error: any }> = [];
+    // If severaProjectId is specified and severaUserId is not, make requests for each opted-in user
+    let allWorkHours: SeveraResponseWorkHours[] = [];
+    if (severaProjectId) {
+      // Request in parallel for each opted-in user
```
Jdallos:
```suggestion
```
Unnecessary comment

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
+            if (endDate) url.searchParams.append("endDate", endDate);
+            const userWorkHours = await api.getWorkHours(url);
+            return userWorkHours;
+          } catch (e) {
+            errors.push({ userGuid: user.guid, error: e });
+            return [];
+          }
+        })
+      );
+      // Combine all results
```
Jdallos:
```suggestion
```
Remove unnecessary comment

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
+      allWorkHours = workHoursResults.flat();
+      // Filtering only for the specified project
```
Jdallos:
```suggestion
      // Filtering only for the specified project
```
Remove unnecessary comment

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
+      allWorkHours = allWorkHours.filter(
+        (workHours) => workHours.project?.guid == severaProjectId
+      );
+    } else {
+      // Standard logic for other cases (by userId or without projectId)
+      // Manually build the URL since buildWorkHoursUrl does not exist
```
Jdallos: Remove unnecessary comment

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
+            // This endpoint/query parameter is working but it is not documented proper in Severa API spec
+            url.searchParams.append("projectGuid", severaProjectId);
+            if (startDate) url.searchParams.append("startDate", startDate);
+            if (endDate) url.searchParams.append("endDate", endDate);
+            const userWorkHours = await api.getWorkHours(url);
+            return userWorkHours;
+          } catch (e) {
+            errors.push({ userGuid: user.guid, error: e });
+            return [];
+          }
```
Jdallos: Could this logic be moved into the sever service as a method there. That's where we want the request logic to be and it keeps the lambda handler as clean and clear as possible

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
+        throw new Error("severaUserId is required when severaProjectId is not specified");
       }
+      if (startDate) url.searchParams.append("startDate", startDate);
+      if (endDate) url.searchParams.append("endDate", endDate);
 
-      const customUrl = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/${endpointPath}`);
-
-      if (startDate) customUrl.searchParams.append("startDate", startDate);
-      if (endDate) customUrl.searchParams.append("endDate", endDate);
+      allWorkHours = await api.getWorkHours(url);
```
Jdallos: Again can this be moved into the severa service as another method.
Also as we no longer have the workhours endpoint request ("severaUserId is required when severaProjectId is not specified"), what happens on the front end? 
I assume it was being requested in some condition from the front end (maybe admin view) and if you are no longer handling that condition I guess it means you plan on changing the UI?

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,56 +11,64 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
-     * will be filtered by project in the Severa API call due to workHours custom url.
-     */
-      if (severaProjectId && severaUserId) {
-        const filteredWorkHoursUserProject = FilterUtilities.filterByUserSevera(workHours.user?.guid, severaUserId);
-        if (!filteredWorkHoursUserProject) return false ;
+    // Filter by opted-in users (just in case there are any "unauthorized" users in the response)
+    const filteredWorkHours = allWorkHours.filter((workHours: SeveraResponseWorkHours) => {
+      if (!workHours.user?.guid || !optInUserGuids.has(workHours.user.guid)) {
+        return false;
       }
```
Jdallos: This is a good idea, but it should be done before the request is made, then we can be sure we are not fetching any non optIn user data.

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -14,23 +14,45 @@ export const getResourceAllocationHandler: APIGatewayProxyHandler = async (event
 
   try {
     const api = CreateSeveraApiService();
+    // Getting the list of opted in users
```
Jdallos: Remove unnecessary comment

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -14,23 +14,45 @@ export const getResourceAllocationHandler: APIGatewayProxyHandler = async (event
...
     const url = buildResourceAllocationUrl(severaUserId);
-    const response = await api.getResourceAllocation(url);
+    let allResourceAllocations: SeveraResponseResourceAllocation[] = [];
+
+    if (url) {
+      // Normal case: for a specific user
+      const response = await api.getResourceAllocation(url);
+      allResourceAllocations = response;
+    } else {
+      // Like in get-filtered-workhours: we collect for all opted-in
```
Jdallos: Remove unnecessary comment

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -14,23 +14,45 @@ export const getResourceAllocationHandler: APIGatewayProxyHandler = async (event
 
   try {
     const api = CreateSeveraApiService();
+    // Getting the list of opted in users
+    const optInUsers = await api.getOptInUsers();
 
     const buildResourceAllocationUrl = (severaUserId?: string) => {
```
Jdallos: Add TSDocs to functions

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -14,23 +14,45 @@ export const getResourceAllocationHandler: APIGatewayProxyHandler = async (event
...
+      // Like in get-filtered-workhours: we collect for all opted-in
+      const resourceAllocationsResults = await Promise.all(
+        optInUsers.map(async (user: { guid: string }) => {
+          try {
+            const userUrl = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/users/${user.guid}/resourceallocations/allocations`);
+            const userResourceAllocations = await api.getResourceAllocation(userUrl);
+            return userResourceAllocations;
+          } catch (e) {
+            // We can add error handling per user if needed
+            return [];
```
Jdallos: We don't necessary need error handling per user, but we do need to log some kind of error message so we know if something is going on. It can remain internal to the lambda so no need to return the errors, just a console.error in here would be fine

src/functions/severa/get-resource-allocations-by-user/handler.ts
```diff
@@ -14,23 +14,45 @@ export const getResourceAllocationHandler: APIGatewayProxyHandler = async (event
...
+          }
+        })
```
Jdallos: Again would it be feesable for this request logic to be moved into an appropriately names severa service method?

Jdallos (changes requested):

src/functions/severa/get-filtered-workhours/handler.ts
StudVladimir: Yes, I will need to change endpoint in Metatavu-home. But requesting data would be the same.

src/functions/severa/get-filtered-workhours/handler.ts
StudVladimir: I want to left it with Set, it's more optimized.
Checking the value of an element is faster and there are no duplicates

src/functions/severa/get-filtered-workhours/handler.ts
```diff
@@ -11,61 +11,32 @@ import type SeveraResponseWorkHours from "src/types/severa/workHour/severaRespon
...
+      allWorkHours = await api.getFilteredWorkHoursForUsers(optInUsers, severaProjectId, startDate, endDate);
+      allWorkHours = allWorkHours.filter((workHours) => workHours.project?.guid == severaProjectId);
     } else {
-      // Standard logic for other cases (by userId or without projectId)
-      // Manually build the URL since buildWorkHoursUrl does not exist
-      let url: URL;
       if (severaUserId) {
-        url = new URL(`${process.env.SEVERA_DEMO_BASE_URL}/v1/users/${severaUserId}/workhours`);
+        if (!optInUserGuids.has(severaUserId)) {
+          allWorkHours = [];
```
Jdallos: Do you think we might want some kind of error message to be returned in this condition?

src/services/severa-api-service.ts
```diff
@@ -348,7 +350,70 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+            const url = new URL(`${baseUrl}/v1/users/${user.guid}/workhours`);
+            // This endpoint/query parameter is working but it is not documented proper in Severa API spec
+            url.searchParams.append("projectGuid", severaProjectId);
+            if (startDate) url.searchParams.append("startDate", startDate);
+            if (endDate) url.searchParams.append("endDate", endDate);
+            const userWorkHours = await (await fetch(url.toString(), {
+              method: "GET",
+              headers: {
+                Authorization: `Bearer ${await getSeveraAccessToken()}`,
+                Client_Id: process.env.SEVERA_DEMO_CLIENT_ID,
```
Jdallos, line 383: Soon we will need to update all references to the demo client id so that it is conditional and in the case of "production environement" it uses the actual client id.
Something to keep in mind before the production update. You could make a new feature branch for this if you have time.

src/services/severa-api-service.ts
```diff
@@ -348,7 +350,70 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+          } catch (e) {
+            errors.push({ userGuid: user.guid, error: e });
+            return [];
+          }
+        })
+      );
+      return workHoursResults.flat();
+    },
+
+    getWorkHoursForUser: async (severaUserId, startDate, endDate) => {
```
Jdallos, line 406: Missing TSDocs

src/services/severa-api-service.ts
```diff
@@ -541,3 +606,61 @@ export const updateSeveraOptInKeyword = async (
...
+ *
+ * @param severaUserId - The GUID of the Severa user (optional).
+ * @param optInUsers - Array of opted-in users ({ guid: string }[]).
+ * @returns {Promise<SeveraResponseResourceAllocation[]>} - Array of resource allocations.
+ */
+export async function getResourceAllocationsByUserOrAll(
+  severaUserId: string | undefined,
+  optInUsers: { guid: string }[]
+): Promise<SeveraResponseResourceAllocation[]> {
+  const api = CreateSeveraApiService();
```
Jdallos, line 630: All of the methods in this file should be within the CreateSeveraApiService, so it's odd this is calling it, and that (after I just checked) there are multiple methods that are not.
This would be another good task, to refactor this file so all methods are within the "createSeveraApiService". Please create a trello card for this and add it (to the top) of the backlog column.

Jdallos: It can be done as a seperate PR and does not need to be done as part of this feature.

Jdallos (approved):
Approved with a few comments

src/services/severa-api-service.ts
Jdallos: Add a card to trello backlog for this task please

src/functions/severa/get-filtered-workhours/handler.ts
StudVladimir: I think it will be good to make notification that user is not optedIn, but I want to test it first, I think that I left output of empty array for UI purposes

StudVladimir: I add error message

PR #182, opened 2025-07-01
Add /users/flextime endpoint for admin dashboard to list opted-in user

AYZBTR:


balashova12 (approved):

StudVladimir (approved):

src/functions/users/flextime/handler.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Lambda handler for listing users with their flextime data.
+ * Only returns users who have opted in to Severa integration.
+ * 
+ * @param event - API Gateway proxy event containing query parameters
+ * @returns Promise resolving to API Gateway proxy result with user flextime data
+ */
+export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) => {
+  try {
+    const keywordId = event.queryStringParameters?.keywordId;
```
Jdallos: We discussed in the spec review that this is not necessary and so this keywordId can be removed from here. We don't need it as we should always be checking users are opted in, so the keywordId is redundant in this context.

src/functions/users/flextime/handler.ts
```diff
@@ -0,0 +1,79 @@
...
+      };
+    }
+    const api = CreateSeveraApiService();
+    const optedInUsers = await api.getOptInUsers();
+    if (!optedInUsers || optedInUsers.length === 0) {
+      return {
+        statusCode: 200,
+        body: JSON.stringify([])
+      };
+    }
```
Jdallos, line 35:
```suggestion
    }
    
```

src/functions/users/flextime/handler.ts
```diff
@@ -0,0 +1,79 @@
...
+    const usersWithFlextime = await Promise.allSettled(
+      optedInUsers.map(async (severaUser) => {
+        try {
+          const flextime = await api.getFlextimeBySeveraUserId(severaUser.guid);
+          return {
```
Jdallos, line 41: Please add a comment here (the object you are returning from within the map) that "this should be typed according to the spec response type for the lambda function". However as we don't have the lambda spec in the lambdas yet (this is a task for someone over the summer) it doesn't need to be done now, but the comment will be helpful in the future.

src/functions/users/flextime/handler.ts
```diff
@@ -0,0 +1,79 @@
...
+                isActive: true
+              }
+            },
+            flextime: {
+              totalFlextimeBalance: flextime?.totalFlextimeBalance || 0,
+              monthFlextimeBalance: flextime?.monthFlextimeBalance || 0
+            }
+          };
+        } catch (error) {
+          throw error;
```
Jdallos: Would be useful to pass a more specific error message and status code from here if possible. Quite sure there are examples of this in other functions in the code.

Jdallos (approved):

src/functions/users/flextime/handler.ts
```diff
@@ -32,34 +22,57 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
     const usersWithFlextime = await Promise.allSettled(
       optedInUsers.map(async (severaUser) => {
         try {
+          /**
+           * TODO: This should be typed according to the spec response type for the lambda function
+           * Spec in the lambdas for the futur
+           */
```
Jdallos: As a note This type of comment with /** is a TSDOc and what we use for function comments. 
For general notes/ TODO we just add a single line comment starting with //
Please update

src/functions/users/flextime/handler.ts
```diff
@@ -32,34 +22,57 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
...
+              attributes: {
+                severaUserId: severaUser.guid,
+                isActive: true
+              }
+            },
+            flextime: null,
+            error: true,
+            /**
+             * Find the appropriate status code for the error
+             */
```
Jdallos: This should be removed, we do not want to return the data like this from the catch as something has gone wrong (which is why we are in the catch). MIght be worth looking into what is typically returned from a catch if this is unclear.

src/functions/users/flextime/handler.ts
```diff
@@ -32,34 +22,57 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
...
               totalFlextimeBalance: flextime?.totalFlextimeBalance || 0,
               monthFlextimeBalance: flextime?.monthFlextimeBalance || 0
             }
           };
         } catch (error) {
-          throw error;
+          /**
+           * Per-user error handling
+           */
+          console.error(`Error fetching flextime for user ${severaUser.guid}:`, error);
```
Jdallos: no need for the console.error here as we are returning the message later in the catch.
Also we don't need TSDOC comments over a console.error and I'd say we don't actually need any comment here.

src/functions/users/flextime/handler.ts
```diff
@@ -32,34 +22,57 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
...
+              firstName: severaUser.firstName || "",
+              lastName: severaUser.lastName || "",
+              attributes: {
+                severaUserId: severaUser.guid,
+                isActive: true
+              }
+            },
+            flextime: null,
+            error: true,
+            /**
```
Jdallos: Remove TSDOc

src/functions/users/flextime/handler.ts
```diff
@@ -32,34 +22,57 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
...
+             * Find the appropriate status code for the error
+             */
+            statusCode: error instanceof Error && (error as any).statusCode ? (error as any).statusCode : 500,
+            message: `Failed to fetch flextime for user ${severaUser.guid}: ${error instanceof Error ? error.message : "Unknown error"}`
```
Jdallos: Mostly right but this should be done in the same structure as the catch return example at the bottom of this file- line 79

src/functions/users/flextime/handler.ts
```diff
@@ -32,34 +22,57 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
...
+          };
         }
       })
     );
-    const successfulResults = usersWithFlextime
-      .filter((result): result is PromiseFulfilledResult<any> => result.status === "fulfilled")
-      .map(result => result.value);
+    /**
+     * Map the results to a more ui format
+     */
```
Jdallos: "more ui format" is not a descriptive comment.
I'd also argue that the original way this was done with the filter and then the map is clearer and so perhaps it should go back to that, unless you disagree?
 const successfulResults = usersWithFlextime
      .filter((result): result is PromiseFulfilledResult<any> => result.status === "fulfilled")
      .map(result => result.value);

src/tests/serverless/mockData.json
```diff
@@ -10,5 +10,6 @@
   "queryStringParameters": {
     "startDate": "2020-11-03",
     "endDate": "2021-11-03"
+
```
Jdallos: Remove unnecessary line break

Jdallos (changes requested):

src/functions/users/flextime/handler.ts
```diff
@@ -22,10 +22,8 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
     const usersWithFlextime = await Promise.allSettled(
       optedInUsers.map(async (severaUser) => {
         try {
-          /**
-           * TODO: This should be typed according to the spec response type for the lambda function
-           * Spec in the lambdas for the futur
-           */
+          //TODO: This should be typed according to the spec response type for the lambda function
+          // Spec in the lambdas for the future
```
Jdallos:
```suggestion
          //TODO: This should be typed according to the response type from the severa spec generated client
```

src/functions/users/flextime/handler.ts
```diff
@@ -39,40 +37,23 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
             }
           };
         } catch (error) {
-          /**
-           * Per-user error handling
-           */
-          console.error(`Error fetching flextime for user ${severaUser.guid}:`, error);
           return {
```
Jdallos, line 54: Here you are returning the body content, see the return type in the catch at the bottom of this file. This is fine but you should also return the status code:
Something like:
{
 statusCode: error instanceof Error && (error as any).statusCode ? (error as any).statusCode : 500,
      body: JSON.stringify({
            code: error instanceof Error && (error as any).statusCode ? (error as any).statusCode : 500,
            message: `Failed to fetch flextime for user ${severaUser.guid}`,
            error: error instanceof Error ? error.message : "Unknown error"
      })
    };
    
    Also try to test locally, do we really need such verbose error parsing?

src/functions/users/flextime/handler.ts
```diff
@@ -39,40 +37,23 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async (event) =>
...
     );
     /**
-     * Map the results to a more ui format
+     * Extract only the successful flextime results from the settled promises
      */
-    const results = usersWithFlextime.map((result) => 
-      result.status === "fulfilled" ? result.value : { error: true, message: "Unexpected promise rejection" }
-    );
+    const successfulResults = usersWithFlextime
+      .filter((result): result is PromiseFulfilledResult<any> => result.status === "fulfilled")
```
Jdallos: Does we have to use "any" here?

Jdallos (changes requested):

src/functions/users/flextime/handler.ts
jessesillman: We can define a new type for user with flextime instead of using "any".

src/functions/users/flextime/handler.ts
```diff
@@ -2,6 +2,20 @@ import type { APIGatewayProxyHandler } from "aws-lambda";
 import { CreateSeveraApiService } from "src/services/severa-api-service";
 import { middyfy } from "src/libs/lambda";
 
+/**
+ * Type representing user with their flextime data.
+ */
+type UserWithFlextime = {
```
Jdallos, line 8: Nice!
Ultimately we will replace this with a type generated from the homeLambdas spec, but that needs to be added to this repo in a future feature

Jdallos (approved):

PR #185, opened 2025-07-10
Feature-184-on-call-setup

StudVladimir:
**To test this function**, I recommend creating a docker container in which to run an instance of DynamoDB.

`sudo docker run -p 8000:8000 amazon/dynamodb-local -jar DynamoDBLocal.jar -inMemory -sharedDb`

`aws configure`

`aws dynamodb create-table   --table-name OnCallSchedule   --attribute-definitions AttributeName=Year,AttributeType=N AttributeName=Week,AttributeType=N   --key-schema AttributeName=Year,KeyType=HASH AttributeName=Week,KeyType=RANGE   --provisioned-throughput ReadCapacityUnits=1,WriteCapacityUnits=1   --endpoint-url http://localhost:8000`

`aws dynamodb create-table   --table-name OnCallNameMap   --attribute-definitions AttributeName=Username,AttributeType=S   --key-schema AttributeName=Username,KeyType=HASH   --provisioned-throughput ReadCapacityUnits=1,WriteCapacityUnits=1   --endpoint-url http://localhost:8000`

`aws dynamodb list-tables --endpoint-url http://localhost:8000`

and change the connection when creating a new instance of DynamoDB

```
const dynamoDb = new DynamoDB.DocumentClient({
        endpoint: "http://localhost:8000",
        region: "local",
        accessKeyId: "fakeMyKeyId",
        secretAccessKey: "fakeSecretAccessKey"
      });
```

### DIFF
--- a/src/functions/on-call/list-on-call-data/handler.ts
+++ b/src/functions/on-call/list-on-call-data/handler.ts
@@ -1,6 +1,7 @@
 import { DynamoDB } from "aws-sdk"
 import { middyfy } from "@libs/lambda";
 import { ValidatedEventAPIGatewayProxyEvent } from "src/libs/api-gateway";
+import { OnCallEntry } from "src/database/models/oncall";
 
 /**
  * Lambda method for loading on-call data
@@ -27,7 +28,7 @@ export const listOnCallDataHandler: ValidatedEventAPIGatewayProxyEvent<any> = as
 
   const dynamoDb = new DynamoDB.DocumentClient();
 
-  // Получаем все записи за указанный год
+  // Get data for the specified year
   const dataResult = await dynamoDb.query({
     TableName: "OnCallSchedule",
     KeyConditionExpression: "#yr = :year",
@@ -39,17 +40,7 @@ export const listOnCallDataHandler: ValidatedEventAPIGatewayProxyEvent<any> = as
     }
   }).promise();
 
-  const data = dataResult.Items || [];
-
-  // Загружаем nameMap из отдельной таблицы
-  const nameMapResult = await dynamoDb.scan({
-    TableName: "OnCallNameMap"
-  }).promise();
-
-  const nameMap: { [key: string]: string } = {};
-  (nameMapResult.Items || []).forEach(item => {
-    nameMap[item.Username] = item.FullName;
-  });
+  const data = (dataResult.Items as OnCallEntry[]) || [];
 
   if (!data.length) {
     return {
@@ -60,9 +51,9 @@ export const listOnCallDataHandler: ValidatedEventAPIGatewayProxyEvent<any> = as
 
   return {
     statusCode: 200,
-    body: JSON.stringify(data.map((entry: any) => ({
+    body: JSON.stringify(data.map((entry) => ({
       ...entry,
-      Person: nameMap[entry.Person] || entry.Person,
+      Person: entry.Person,
       Paid: entry.Paid || false
     })))
   };
--- a/src/functions/on-call/list-on-call-data/handler.ts
+++ b/src/functions/on-call/list-on-call-data/handler.ts
@@ -1,8 +1,5 @@
-import { S3 } from "aws-sdk"
-import { OnCallEntry, PaidData } from "../../../types/on-call";
-import S3Utils from "@libs/s3-utils";
+import { DynamoDB } from "aws-sdk"
 import { middyfy } from "@libs/lambda";
-import Config from "../../../app/config";
 import { ValidatedEventAPIGatewayProxyEvent } from "src/libs/api-gateway";
 
 /**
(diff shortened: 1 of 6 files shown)


src/functions/on-call/list-on-call-data/handler.ts
```diff
@@ -28,18 +25,33 @@ export const listOnCallDataHandler: ValidatedEventAPIGatewayProxyEvent<any> = as
     }
   }
 
-  const s3 = new S3();
+  const dynamoDb = new DynamoDB.DocumentClient();
 
-  const nameMapFile = "name-map.json";
-  const yearFile = `${year}.json`;
-  const paidFile = "paid.json";
+  // Получаем все записи за указанный год
```
Jdallos: Remove comments

src/functions/on-call/list-on-call-data/handler.ts
```diff
@@ -28,18 +25,33 @@ export const listOnCallDataHandler: ValidatedEventAPIGatewayProxyEvent<any> = as
...
+      "#yr": "Year"
+    },
+    ExpressionAttributeValues: {
+      ":year": year
+    }
+  }).promise();
+
+  const data = dataResult.Items || [];
+
+  // Загружаем nameMap из отдельной таблицы
```
Jdallos: Remove comment

src/functions/on-call/list-on-call-data/handler.ts
```diff
@@ -28,18 +25,33 @@ export const listOnCallDataHandler: ValidatedEventAPIGatewayProxyEvent<any> = as
...
-  const nameMap = await S3Utils.loadJson<{ [key: string]: string }>(s3, bucket, nameMapFile) || {};
-  const data = await S3Utils.loadJson<OnCallEntry[]>(s3, bucket, yearFile);
-  const paidData = await S3Utils.loadJson<PaidData>(s3, bucket, paidFile) || {};
+  const nameMap: { [key: string]: string } = {};
+  (nameMapResult.Items || []).forEach(item => {
+    nameMap[item.Username] = item.FullName;
+  });
 
-  if (!data || !nameMap || !paidData) {
+  if (!data.length) {
```
Jdallos, line 45: If you are only checking data.length then this could be moved to just after data is declared (before the nameMapResult scan)

src/functions/on-call/list-on-call-data/handler.ts
```diff
@@ -28,18 +25,33 @@ export const listOnCallDataHandler: ValidatedEventAPIGatewayProxyEvent<any> = as
...
   }
 
-  const s3 = new S3();
+  const dynamoDb = new DynamoDB.DocumentClient();
 
-  const nameMapFile = "name-map.json";
-  const yearFile = `${year}.json`;
-  const paidFile = "paid.json";
+  // Получаем все записи за указанный год
+  const dataResult = await dynamoDb.query({
```
Jdallos, line 32: Can you explain why we now need 2 data base tables for listing the on call data? If all we need is the persons name, paid boolean and I guess dates- it seems one table should be fine.

As far as I can tell you are also missing a model (see database/models) to describe the on call data.
It would be good if we had a type/ model to describe the dataResult and nameMapResult as currently they are both typed as any

src/functions/on-call/update-paid/handler.ts
```diff
@@ -25,19 +23,25 @@ export const updatePaidHandler: ValidatedEventAPIGatewayProxyEvent<any> = async
...
-  const bucket = Config.get().onCall.bucketName;
-  const paidFile = "paid.json";
-
-  const paidData = await S3Utils.loadJson<PaidData>(s3, bucket, paidFile) || {};
-  if (!paidData[year]) paidData[year] = {}
-  paidData[year][week] = paid;
-  await S3Utils.saveJson(s3, bucket, paidFile, paidData);
+  const dynamoDb = new DynamoDB.DocumentClient();
+
+  // Update only after Set Paid for record direct Year adn Week
```
Jdallos: This comment should be removed or clarified

src/functions/on-call/weekly-check/handler.ts
```diff
@@ -54,22 +53,20 @@ export const weeklyCheckHandler : ValidatedEventAPIGatewayProxyEvent<any> = asyn
...
-  } else {
-    yearJson.push({
-      Week: nextWeek.week,
-      Person: nextWeek.user
-    });
-  }
-
-  await S3Utils.saveJson(s3, bucket, yearFile, yearJson);
+  // Update or create a new record in DynamoDB
+  await dynamoDb.put({
```
Jdallos, line 70: Again we really need a model to describe the structure of the data here- software-registry/ create-software might be a useful reference

src/functions/on-call/weekly-check/index.ts
```diff
@@ -4,9 +4,17 @@ const { ONCALL_WEEKLY_SCHEDULE_TIMER } = process.env;
...
+    ] : []),
     {
-      schedule: ONCALL_WEEKLY_SCHEDULE_TIMER,
+      httpApi: {
+        path: "/on-call/weekly-check",
+        method: "get"
+      }
     }
-  ] : []
+  ]
```
Jdallos: The original idea was we are fetching the data from splunk and then updating out s3 (now DB), the cron timer was configured to be once a week as it's just a single row update. We didn't need to call this function as an API endpoint previously and I don't really see any reason to now. So I'd suggest reverting this to remove the httpApi config. 

Although it doesn't really do any harm I think in the long run it keeps the spec free of unnecessary endpoints and clutter (it's already pretty cluttered).

Jdallos (changes requested):
Requested changes, mostly want some clarification about the need for 2 DB tables.

src/functions/on-call/list-on-call-data/handler.ts
StudVladimir: I left now only one table.

Jdallos (approved):

PR #188, opened 2025-08-21
Feature-187-refactor-severa-api-service

StudVladimir:
Refactor CreateSeveraApiService.

Put next functions to interface and put implementation inside CreateSeveraApiService.

getResourceAllocationsByUserOrAll()
checkKeywordExists()
fetchUserByEmail()
getUserKeywords()
updateSeveraOptInKeyword()

### DIFF
--- a/src/services/severa-api-service.ts
+++ b/src/services/severa-api-service.ts
@@ -681,7 +681,10 @@ const getSeveraAccessToken = async (): Promise<string> => {
 
 // Utility function to get correct Severa client ID based on environment
 const getSeveraClientId = (): string => {
-  return process.env.NODE_ENV === "production"
-    ? process.env.SEVERA_CLIENT_ID
-    : process.env.SEVERA_DEMO_CLIENT_ID;
+  switch (process.env.STAGE) {
+    case "production":
+      return process.env.SEVERA_CLIENT_ID;
+    default:
+      return process.env.SEVERA_DEMO_CLIENT_ID;
+  }
 };
--- a/src/services/severa-api-service.ts
+++ b/src/services/severa-api-service.ts
@@ -28,6 +28,10 @@ export interface SeveraApiService {
   getFilteredWorkHoursForUsers: (users: { guid: string }[], severaProjectId: string, startDate?: string, endDate?: string) => Promise<SeveraResponseWorkHours[]>;
   getWorkHoursForUser: (severaUserId: string, startDate?: string, endDate?: string) => Promise<SeveraResponseWorkHours[]>;
   getResourceAllocationsByUserOrAll: (severaUserId: string | undefined, optInUsers: { guid: string }[]) => Promise<SeveraResponseResourceAllocation[]>;
+  checkKeywordExists: (keyword: string) => Promise<any>;
+  fetchUserByEmail: (email: string) => Promise<any>;
+  getUserKeywords: (userGuid: string) => Promise<any>;
+  updateSeveraOptInKeyword: (userGuid: string, isSeveraOptIn: string, isSeveraOptInKeywordGuid: string) => Promise<any>;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


jessesillman (approved):

balashova12 (approved):

src/services/severa-api-service.ts
```diff
@@ -27,6 +27,11 @@ export interface SeveraApiService {
   getPreviousWeekHours: (severaUserId: string) => Promise<SeveraResponsePreviousWorkHours[]>;
   getFilteredWorkHoursForUsers: (users: { guid: string }[], severaProjectId: string, startDate?: string, endDate?: string) => Promise<SeveraResponseWorkHours[]>;
   getWorkHoursForUser: (severaUserId: string, startDate?: string, endDate?: string) => Promise<SeveraResponseWorkHours[]>;
+  getResourceAllocationsByUserOrAll: (severaUserId: string | undefined, optInUsers: { guid: string }[]) => Promise<SeveraResponseResourceAllocation[]>;
+  checkKeywordExists: (keyword: string) => Promise<any>;
+  fetchUserByEmail: (email: string) => Promise<any>;
+  getUserKeywords: (userGuid: string) => Promise<any>;
+  updateSeveraOptInKeyword: (userGuid: string, isSeveraOptIn: string, isSeveraOptInKeywordGuid: string) => Promise<any>;
```
Jdallos: Lets avoid typing the return as any, we should be able to get the response type from the severa client, which it looks like is already set up in this project. There is also the use of SeveraResponse types e.g. SeveraResponseWorkhours which are custom types created by us, but again these could be replaced by the types generated from the client.

Considering this I think it would be reasonable to make sure the severa API spec is generated (ideally just the models rather than the Apis) and then this could also be committed as part of this PR.

src/services/severa-api-service.ts
```diff
@@ -421,7 +426,213 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+
+      const url = buildResourceAllocationUrl(severaUserId);
+      let allResourceAllocations: SeveraResponseResourceAllocation[] = [];
+
+      if (url) {
+        const response = await (await fetch(url.toString(), {
+          method: "GET",
+          headers: {
+            Authorization: `Bearer ${await getSeveraAccessToken()}`,
+            Client_Id: process.env.SEVERA_DEMO_CLIENT_ID,
```
Jdallos: Could this also be a good PR to refactor the SEVERA_DEMO_CLIENT_ID env, we should add a SEVERA_CLIENT_ID env, both should be optional (and exclusive) so one must be present, both will not be.
Locally we will only be using the demo client but for the production we will add the production id (and not the demo id)

Jdallos (changes requested):

src/services/severa-api-service.ts
Jdallos: This update should be done everywheere the demo client is currently used to make sure it will work when we switch to production environemnt

src/services/severa-api-service.ts
```diff
@@ -675,3 +678,10 @@ const getSeveraAccessToken = async (): Promise<string> => {
     throw new Error(`Failed to get Severa access token: ${error.message}`);
   }
 };
+
+// Utility function to get correct Severa client ID based on environment
+const getSeveraClientId = (): string => {
+  return process.env.NODE_ENV === "production"
```
Jdallos: I'm not certain that this will work as intended, as we have the staging environment which might also be classed as NODE_ENV === "production.
So we have 2 options, we handle this based on what envs are available in the pipeline, so we ensure only demo is in the staging pipeline and not demo id is in the production. This could work with a conditional selector for whichever one is truthy.

Doing at you are trying to do would be better, if we can confirm it is production in production but not in the staging env, perhaps the best way to check this is to add a log so we can see what the NODE ENV is when it's deployed to staging (vie develop merge).

Open to alternative suggestions

Jdallos (approved):

src/services/severa-api-service.ts
```diff
@@ -681,7 +681,9 @@ const getSeveraAccessToken = async (): Promise<string> => {
 
 // Utility function to get correct Severa client ID based on environment
 const getSeveraClientId = (): string => {
-  return process.env.NODE_ENV === "production"
-    ? process.env.SEVERA_CLIENT_ID
-    : process.env.SEVERA_DEMO_CLIENT_ID;
+  if (process.env.STAGE === "production") {
```
Jdallos, line 684: This is fine for now as we might end up changing this later to have a single env, and set it as appropriate depending on pipeline. 
Something to figure out during production review (or after)

Jdallos (approved):

PR #214, opened 2025-09-08
feature-194-node-update

StudVladimir:


JuusoHeikkinen02 (approved):

jessesillman (approved):

src/database/client/index.ts
```diff
@@ -1,20 +1,20 @@
...
  */
-const createDynamoDBClient = (): DocumentClient => {
-  if (process.env.IS_OFFLINE) {
-    return new AWS.DynamoDB.DocumentClient({
-      region: "localhost",
-      endpoint: "http://localhost:8000",
-    });
-  }
+const createDynamoDBClient = (): DynamoDBDocumentClient => {
+  const client = process.env.IS_OFFLINE
```
Jdallos, line 10: Does doesn't seem related to the node update.
I'd rather we testted the serverless offline plugins before setting up alternative "offline" methods. Or is this doing something else?

src/database/services/articles-api-service.ts
```diff
@@ -70,12 +71,11 @@ class ArticlesApiService {
    * @returns article entry or null if not found
    */
   public findArticleById = async(id: string): Promise<ArticleModel | null> => {
-    const articles = await this.docClient
-      .get({
-        TableName: TABLE_NAME,
-        Key: { id: id },
-      }).promise();
-
+    const params: GetCommandInput = {
```
Jdallos, line 74: Again, were these changes were required for the node update?

src/database/services/software-service.ts
```diff
@@ -30,10 +30,11 @@ class SoftwareService {
       lastUpdatedAt: new Date().toISOString(),
     };
     try {
-      await this.docClient.put({
+      const params: PutCommandInput = {
         TableName: tableName,
         Item: newSoftware,
-      }).promise();
+      };
+      await this.docClient.send(new PutCommand(params));
```
Jdallos, line 37: Again more DB request changes, I want to confirm these were required by the node update and not just refactorings (which I would have preferred to do seperately to allow us to more easily debug any issues that come up due to node update)

src/libs/s3-utils.ts
```diff
@@ -1,53 +0,0 @@
-import { S3 } from "aws-sdk"
-
-/**
- * Set of S3 utilities
- */
-namespace S3Utils {
```
Jdallos, line 6: Are you certain this file is now unused. We no longer use the s3 for on call but it is used in articles.

StudVladimir: This utility has not been used anywhere. Articles use DynamoDB.

Jdallos: Yes, articles also use s3 for image storage. However as long as nothing in this namespace was used for that then removal is fine.

Jdallos (approved):

PR #215, opened 2025-09-15
Feature-207-add-on-call-data-from-json

StudVladimir:


JuusoHeikkinen02 (approved):

jessesillman (approved):

src/functions/on-call/create-on-call-data-from-json/handler.ts
```diff
@@ -0,0 +1,77 @@
...
+import { middyfy } from "@libs/lambda";
+import type { OnCallImportEntry, OnCallImportError } from "src/database/models/oncall";
+
+/**
+ * Lambda method for importing on-call data from JSON
+ *
+ * Expects body to be an array of objects with "Week" (number) and "Person" (string) properties
+ * and a query parameter "year" (number)
+ */
+export const onCallImportFromJsonHandler = async (event) => {
```
Jdallos, line 13: Looks good, only change I would suggest is we make sure the requesting user (via token) has admin credentials. We already have this in the utils, I think Juuso refactored it today so might be worth checking usage with him.

Jdallos (changes requested):

src/functions/on-call/create-on-call-data-from-json/handler.ts
```diff
@@ -0,0 +1,88 @@
...
+  const token = event.headers?.Authorization?.split(' ')[1];
+  const decodedJWT = jwt.decode(token);
+  const isAdmin = decodedJWT?.realm_access?.roles?.includes("admin") || false;
+  if (!isAdmin) {
+    return {
+      statusCode: 403,
+      body: "Forbidden: admin access required",
+    };
+  }
```
Jdallos: Can you not use the reusable utility function that does this that Juuso set up, it's in:
/src/libs/auth-utils.ts

Jdallos (approved):

PR #232, opened 2025-10-01
bugfix-231-fix-authorization-failure

JuusoHeikkinen02:
Fixed the isAdminUser functions failure that was caused by not returning the needed data from the token.


src/libs/auth-utils.ts
```diff
@@ -44,7 +44,8 @@ export const getAuthDataFromToken = (event: { headers: APIGatewayProxyEvent['hea
     console.log('Decoded JWT Token:', JSON.stringify(decodedToken, null, 2));
```
Jdallos: Remove this log

Jdallos (approved):

PR #245, opened 2025-10-13
bugfix-244-restore-functionality-to-questionnaire-screen

Alexander-K-byte:


src/functions/questionnaire/create-questionnaire/handler.ts
```diff
@@ -44,7 +44,7 @@ export const createQuestionnaireHandler: ValidatedEventAPIGatewayProxyEvent<Ques
   }
 
   const newQuestionnaireId: string = uuidv4();
-  let questionnaireResponse: Questionnaire | undefined;
+  let questionnaireResponse: QuestionnaireModel | undefined;
```
Jdallos: Should there be some parsing here as we actually want to return what we say we will return in the Spec, not based on the model.

Jdallos: So Questionnaire not QuestionnaireModel

Jdallos (changes requested):

src/functions/questionnaire/create-questionnaire/handler.ts
Alexander-K-byte: Yeah, I just copy pasted from the original file. Sorted imports and replaced with Questionnaire

Jdallos (approved):

PR #246, opened 2025-10-14
bugfix-244-restore-functionality-to-questionnaire-screen

Alexander-K-byte:
reverted create questionnaire and modified to use appropriate model, update software required some minimal changes also.

Both tested locally and appear to be working as expected, further testing of other functions required,

### DIFF
--- a/src/functions/software-registry/update-software/handler.ts
+++ b/src/functions/software-registry/update-software/handler.ts
@@ -18,7 +18,6 @@ const softwareService = new SoftwareService(docClient);
 export const updateSoftwareHandler: ValidatedEventAPIGatewayProxyEvent<SoftwareRegistry> = async (
   event
 ) => {
-  console.log("Received event:", JSON.stringify(event));
   try {
     if (!isAdminUser(event)) {
       return {
@@ -30,7 +29,6 @@ export const updateSoftwareHandler: ValidatedEventAPIGatewayProxyEvent<SoftwareR
       };
     }
     const { id } = event.pathParameters || {};
-    console.log("Path parameter (id):", id);
 
     if (!id) {
       return {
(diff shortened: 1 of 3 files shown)


src/functions/software-registry/update-software/handler.ts
```diff
@@ -20,7 +18,7 @@ const softwareService = new SoftwareService(docClient);
 export const updateSoftwareHandler: ValidatedEventAPIGatewayProxyEvent<SoftwareRegistry> = async (
   event
 ) => {
-
+  console.log("Received event:", JSON.stringify(event));
```
Jdallos: Do you deliberately want to leave this log in, I will need to access the logs through AWS if you want to use this for debugging? Otherwise remove

src/functions/software-registry/update-software/handler.ts
```diff
@@ -31,36 +29,49 @@ export const updateSoftwareHandler: ValidatedEventAPIGatewayProxyEvent<SoftwareR
         })
       };
     }
-
     const { id } = event.pathParameters || {};
+    console.log("Path parameter (id):", id);
```
Jdallos: Same here, all console.logs should be removed unless they are deliberately here for temporary debugging on deployment in which case they should be commented as such

src/functions/software-registry/update-software/handler.ts
```diff
@@ -31,36 +29,49 @@ export const updateSoftwareHandler: ValidatedEventAPIGatewayProxyEvent<SoftwareR
...
     }
 
-    const data: SoftwareRegistry =
-      typeof event.body === "string" ? JSON.parse(event.body) : event.body;
+    let data: SoftwareRegistry;
+    if (typeof event.body === "string") {
+      data = JSON.parse(event.body);
+    } else {
+      data = event.body as SoftwareRegistry;
+    }
```
Jdallos: Wasn't the ternary just fine before?, you can also do the casting in the ternary which would work fine

Jdallos (approved):

src/functions/software-registry/update-software/handler.ts
Alexander-K-byte: Was mistakenly added when reverting code

src/functions/software-registry/update-software/handler.ts
Alexander-K-byte: same as previous

JuusoHeikkinen02 (approved):

PR #249, opened 2025-10-16
bugfix-248-draft-vacation-notifying-admin

JuusoHeikkinen02:
Made if checks to create/delete/update handlers so admin does not get notified if draft request is created or deleted. Admin gets the notification after the user updates the request to not be draft anymore

### DIFF
--- a/src/functions/vacation-request/update-vacation-request/handler.ts
+++ b/src/functions/vacation-request/update-vacation-request/handler.ts
@@ -3,7 +3,10 @@ import { middyfy } from "@libs/lambda";
 import { vacationRequestService } from "src/database/services";
 import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
 import type { VacationRequest } from "src/generated/homeLambdasModels/model/vacationRequest";
-import { notifyUserVacationStatusUpdatedAll } from "src/notifications/vacation-notifications";
+import {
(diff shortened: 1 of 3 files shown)


Alexander-K-byte (approved):

jessesillman (approved):

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -93,6 +97,15 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
         updatedStatus: currentStatus
       });
     }
+    if (draftStatusChanged) {
```
Jdallos, line 100: Could a user have a vacation that has been submitted and then change it to a draft? In which case this would send a draft to an admin right?

Probably we should ensure that it is not possible to change a submitted vacation to draft. Please check, if not lets create a card in the backlog for this.

Jdallos (approved):

src/functions/vacation-request/update-vacation-request/handler.ts
JuusoHeikkinen02: Its not possible for user to change submitted request back to draft through the UI. The draft state is only updated trough creating new draft request or removing the draft state with the submit to review button.

PR #252, opened 2025-10-20
Feature 247 vacations consume user vaction days

JuusoHeikkinen02:
Added util file that splits vacation days so it consumes the correct years vacation days if the vacation happens around year shifts. Added function that consumes users unspent vacationdays based on the days on the approved request.

User gets an error if trying to create a new request if it exceeds the unspentDays amount, Same check is made for the approval of a request just in case so it does not update users unspentdays if the request consumes more days.

### DIFF
--- a/src/libs/vacation-utils.ts
+++ b/src/libs/vacation-utils.ts
@@ -50,23 +50,38 @@ export const splitVacationDaysByYear = (
 export const updateRemainingVacationDays = async (
   userId: string,
   daysToSubtract: number,
+  status: string,
   year: string
-) => {
+): Promise<boolean> => {
   const keycloakApiService = CreateKeycloakApiService();
-  const userDetails = await keycloakApiService.getUserAttributes(userId);
-  const unspentArr = userDetails.unspentVacationDaysByYear || [];
-  const index = unspentArr.findIndex((s) => s.startsWith(`${year}:`));
-  const currentValueStr = index >= 0 ? unspentArr[index].split(":")[1] : "0";
-  const currentValue = Number(currentValueStr);
+  const user = await keycloakApiService.getUserAttributes(userId);
+
+  user.attributes = user.attributes || [];
+
+  const unspentVacationDaysByYear = user.unspentVacationDaysByYear || [];
+
+  const currentEntry = unspentVacationDaysByYear.find((s) => s.startsWith(`${year}:`));
+  const currentValue = currentEntry ? Number(currentEntry.split(":")[1]) : 0;
+
+  if (currentValue < daysToSubtract) {
+    return false;
+  }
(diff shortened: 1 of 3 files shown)


jessesillman (approved):

src/libs/vacation-utils.ts
```diff
@@ -0,0 +1,86 @@
...
+export const updateRemainingVacationDays = async (
+  userId: string,
+  daysToSubtract: number,
+  status: string,
+  year: string
+): Promise<boolean> => {
+  const keycloakApiService = CreateKeycloakApiService();
+  const user = await keycloakApiService.getUserAttributes(userId);
+
+  user.attributes = user.attributes || [];
```
MarcelMeta: Is user.attributes used anywhere?

MarcelMeta (approved):

src/libs/vacation-utils.ts
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * Counts weekdays (Monday to Friday) by default, and optionally includes Saturdays
+ * if the vacation period is longer than 7 days.
+ *
+ * @param startDate - The start date of the vacation in ISO format (YYYY-MM-DD).
+ * @param endDate - The end date of the vacation in ISO format (YYYY-MM-DD).
+ *
+ * @returns An object where the keys are years and the values are the number of vacation days in that year.
+ */
+export const splitVacationDaysByYear = (
```
Jdallos, line 17: This will require some additional logic to handle part time workers vacations, however as we don't have info from severa yet whether a worker is part time or not, we can't really implement it yet.

Leave a TODO: Comment that this requires updating to handle part time workers.

src/libs/vacation-utils.ts
```diff
@@ -0,0 +1,84 @@
...
+  const remainingDays = Math.max(currentValue - daysToSubtract, 0);
+  const formattedValue = `${year}:${String(remainingDays).padStart(3, "0")}`;
+
+  const updatedUnspent = [...unspentVacationDaysByYear];
+  const existingIndex = updatedUnspent.findIndex((s) => s.startsWith(`${year}:`));
+
+  if (existingIndex >= 0) {
+    updatedUnspent[existingIndex] = formattedValue;
+  } else {
+    updatedUnspent.push(formattedValue);
```
Jdallos, line 78: This is the condition when its the first entry of a new year, right?

Jdallos (approved):

src/libs/vacation-utils.ts
JuusoHeikkinen02: yes if there is only one index for years for example "2025: 021" it updates it, but the upper check updates all of the indexes with their appropriate data.

PR #259, opened 2025-10-28
feature-258-Find-users-contracted-work-week

Alexander-K-byte:
Created handler to find users normal work week, returns numbered array of normal days worked; added fields to types to allow date to be parsed and whether day falls on holiday; modified severa api service to allow optional date ranges to search; index file created with path of severa/user/workWeek.

### DIFF
--- a/src/functions/severa/get-filtered-workdays/handler.ts
+++ b/src/functions/severa/get-filtered-workdays/handler.ts
@@ -1,28 +1,42 @@
 import type { APIGatewayProxyHandler } from "aws-lambda";
 import { DateTime } from "luxon";
+import Config from "src/app/config";
 import { middyfy } from "src/libs/lambda";
 import { CreateSeveraApiService } from "src/services/severa-api-service";
 import type SeveraResponseWorkDays from "src/types/severa/workDays/severaResponseWorkDays";
 
 /**
- * Lambda handler to calculate the contracted work week of a user from Severa.
- * Get logged-in user's email from the authorizer, then fetch Severa GUID for user.
+ * Lambda handler to calculate the contracted work week of a user from Severa. Check user has SeveraId which should come from opting in.
+ * Tests email matches requested user data so url cannot be manipulated.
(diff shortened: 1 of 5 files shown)


MarcelMeta (approved):

JuusoHeikkinen02 (approved):

src/functions/severa/get-filtered-workdays/handler.ts
JuusoHeikkinen02, line 1: Should there be a check for if the user is OptIn. Other handlers that fetch user data have it.

jessesillman (approved):

src/functions/severa/get-filtered-workdays/handler.ts
Alexander-K-byte: Yeah, will try to re-write this to reflect the flextime handler, it has nice simple logic that I should have copied.

Alexander-K-byte: Modified this so that it checks if the users has SeveraUserId from the request path, which (I could be wrong) comes from opting in to Severa, added email check to ensure logged in email matches requested Severa GUID email so that users can't manipulate the URL to query the endpoint for someone elses data

src/functions/severa/get-filtered-workdays/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "get",
+        path: "severa/users/workWeek",
```
JuusoHeikkinen02: The path needs the parameter for userId like for example severa/users/{userId}/workweek 
because otherwise it will be undefined when calling it in the handler  const severaUserId = event.pathParameters?.userId;

Alexander-K-byte:
Closed this PR as it was made to develop and not temp-develop

PR #266, opened 2025-11-03
Feature-258-get-users-contracted-work-week

Alexander-K-byte:
Use workdays endpoint, calculate a normal working week for the logged in user, returns numbered array of days worked and also total work days in week.

### DIFF
--- a/src/functions/severa/get-filtered-workdays/handler.ts
+++ b/src/functions/severa/get-filtered-workdays/handler.ts
@@ -1,28 +1,42 @@
 import type { APIGatewayProxyHandler } from "aws-lambda";
 import { DateTime } from "luxon";
+import Config from "src/app/config";
 import { middyfy } from "src/libs/lambda";
 import { CreateSeveraApiService } from "src/services/severa-api-service";
 import type SeveraResponseWorkDays from "src/types/severa/workDays/severaResponseWorkDays";
 
 /**
- * Lambda handler to calculate the contracted work week of a user from Severa.
- * Get logged-in user's email from the authorizer, then fetch Severa GUID for user.
+ * Lambda handler to calculate the contracted work week of a user from Severa. Check user has SeveraId which should come from opting in.
+ * Tests email matches requested user data so url cannot be manipulated.
  * Lastly find a holiday free week for the user to find a normal expecetd work week.
  */
 export const getContractedWorkWeekHandler: APIGatewayProxyHandler = async (event) => {
-  try {
-    const api = CreateSeveraApiService();
+  const severaUserId = event.pathParameters?.userId;
 
-    const userEmail = event.requestContext.authorizer.claims.email;
-    if (!userEmail) {
+  try {
+    if (!severaUserId) {
       return {
         statusCode: 400,
-        body: JSON.stringify({ message: "User email not found" })
+        body: JSON.stringify({
+          message: "You need to opt in with the Severa service, Severa User ID required"
+        })
       };
     }
 
-    const user = await api.fetchUserByEmail(userEmail);
-    const userGuid = user.guid;
+    const api = CreateSeveraApiService();
+
+    const severaUser = await api.getUser(severaUserId);
+
+    const testUserEmail = Config.get().testUser.email;
+    const keycloakEmail = testUserEmail || event.requestContext.authorizer.claims.email;
+
+    if (severaUser.email !== keycloakEmail) {
+      return {
+        statusCode: 403,
+        body: JSON.stringify({ message: "You can only view your own information" })
+      };
+    }
+
     const today = DateTime.now();
     let contractedWeek: number[] = [];
     let foundWeek = false;
@@ -32,7 +46,7 @@ export const getContractedWorkWeekHandler: APIGatewayProxyHandler = async (event
       const weekEnd = weekStart.plus({ days: 6 });
 
       const workWeekData: SeveraResponseWorkDays[] = await api.getWorkWeek(
-        userGuid,
+        severaUserId,
         weekStart.toISODate(),
         weekEnd.toISODate()
       );
@@ -48,13 +62,6 @@ export const getContractedWorkWeekHandler: APIGatewayProxyHandler = async (event
         contractedWeek = workdays.map((day) => DateTime.fromISO(day.date).weekday % 7);
         foundWeek = true;
       }
-    }
-
-    if (!foundWeek) {
-      return {
-        statusCode: 404,
-        body: JSON.stringify({ message: "No holiday-free week found in recent weeks" })
-      };
     }
 
     return {
--- a/src/functions/severa/get-filtered-workdays/handler.ts
+++ b/src/functions/severa/get-filtered-workdays/handler.ts
@@ -52,13 +52,17 @@ export const getContractedWorkWeekHandler: APIGatewayProxyHandler = async (event
       );
 
       /**
-       * Filter out zero hour days and holidays
-       *
-       * @return a numeric array of normal work days in a week for the user where Monday = 1...Sunday = 7, and number of active workdays in a week.
+       * foundWeek = previous 4 weeks where week contains no holidays
+       * When week is found, find workdays where expected hours > 0
        */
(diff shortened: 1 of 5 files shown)


src/functions/severa/get-filtered-workdays/handler.ts
```diff
@@ -0,0 +1,82 @@
...
+        })
+      };
+    }
+
+    const api = CreateSeveraApiService();
+
+    const severaUser = await api.getUser(severaUserId);
+
+    const testUserEmail = Config.get().testUser.email;
+    const keycloakEmail = testUserEmail || event.requestContext.authorizer.claims.email;
```
MarcelMeta, line 30: This and other "test data" will probably change when going on production?

src/functions/severa/get-filtered-workdays/handler.ts
```diff
@@ -0,0 +1,82 @@
...
+      const workWeekData: SeveraResponseWorkDays[] = await api.getWorkWeek(
+        severaUserId,
+        weekStart.toISODate(),
+        weekEnd.toISODate()
+      );
+
+      /**
+       * Filter out zero hour days and holidays
+       *
+       * @return a numeric array of normal work days in a week for the user where Monday = 1...Sunday = 7, and number of active workdays in a week.
```
MarcelMeta: I think we should stick with monday=1 ... sunday=0.

MarcelMeta (approved):

src/functions/severa/get-filtered-workdays/handler.ts
Alexander-K-byte: Most likely, although we are currently pushing this to the temp develop branch, which will then merge to develop after production is merged. So it could be quite a while before this ever goes to production.

src/functions/severa/get-filtered-workdays/handler.ts
Alexander-K-byte: I decided to go with the default of Monday = 1 and Sunday = 7 since the calculations that are in place for vacation days consumption assume that this is the data that is incoming and converts the default data with a workday - 1 calculation so Monday is then zero and Sunday = 6. It might require rewriting a few blocks of code to deal with that change in the time utils file.

JuusoHeikkinen02 (approved):

src/functions/severa/get-filtered-workdays/handler.ts
```diff
@@ -0,0 +1,86 @@
...
+import { DateTime } from "luxon";
+import Config from "src/app/config";
+import { middyfy } from "src/libs/lambda";
+import { CreateSeveraApiService } from "src/services/severa-api-service";
+import type SeveraResponseWorkDays from "src/types/severa/workDays/severaResponseWorkDays";
+
+/**
+ * Lambda handler to calculate the contracted work week of a user from Severa. Check user has SeveraId which should come from opting in.
+ * Tests email matches requested user data so url cannot be manipulated.
+ * Lastly find a holiday free week for the user to find a normal expecetd work week.
```
Jdallos: This is more detail than typical in TSDocs but as this is is trainee project it's fine- try to keep docs shorter where possible so people can (and will) read at a glance

Jdallos:
```suggestion
 * Tests email matches requested user data so url cannot be manipulated.
 * Lastly find a holiday free week for the user to find a normal expeceted work week.
```

src/functions/severa/get-filtered-workdays/handler.ts
```diff
@@ -0,0 +1,86 @@
...
+        statusCode: 403,
+        body: JSON.stringify({ message: "You can only view your own information" })
+      };
+    }
+
+    const today = DateTime.now();
+    let contractedWeek: number[] = [];
+    let foundWeek = false;
+
+    for (let i = 0; i < 4 && !foundWeek; i++) {
```
Jdallos: why 4? Perhaps this could be a named variable to increase clarity

src/functions/severa/get-filtered-workdays/handler.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (!severaUserId) {
+      return {
+        statusCode: 401,
+        body: JSON.stringify({
+          message: "You need to opt in with the Severa service, Severa User ID required"
+        })
+      };
+    }
+
+    const api = CreateSeveraApiService();
```
Jdallos: calling this severaApi rather than just api would improve clarity as we are dealing with multiple 3rd party API's

src/functions/severa/get-filtered-workdays/handler.ts
Jdallos: If it's that you are checking 4 weeks how did you determine that's a suitable amount to check?

src/functions/severa/get-filtered-workdays/handler.ts
```diff
@@ -0,0 +1,86 @@
...
+      const hasHoliday = workWeekData.some((day) => day.isHoliday);
+      if (hasHoliday) {
+        continue;
+      }
+
+      const workdays = workWeekData.filter((day) => day.expectedHours > 0);
+
+      if (workdays.length > 0) {
+        contractedWeek = workdays.map((day) => DateTime.fromISO(day.date).weekday);
+        foundWeek = true;
```
Jdallos: Could you break here rather than set foundWeek to true- would be cleaner and more explicit

src/functions/severa/get-filtered-workdays/handler.ts
Jdallos: Is it that 4 weeks is enough to guarantee that you get a week with no holidays?

src/services/severa-api-service.ts
```diff
@@ -169,34 +196,40 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
-     * 
-     * @param severaUserId Severa user id
-     * @returns Workdays of a user
+     * Fetches workdays for a user from Severa within an optional data range
+     * Modified for calculating user vacation days consumption
+     *
+     * @param severaUserId - GUID of the Severa user whos workdays are being retrieved
+     * @param startDate - Optional start date for filtering workdays in ISO format, defaults to 7 days before
+     * @param endDate - Optional end date, defaults to today
+     * @returns - {Promise<SeveraResponseWorkDays[]>} - A promise that resolves to an array of workday objects for the user
```
Jdallos:
```suggestion
     * @returns - {Promise<SeveraResponseWorkDays[]>}
```

src/services/severa-api-service.ts
```diff
@@ -169,34 +196,40 @@ export const CreateSeveraApiService = (): SeveraApiService => {
     },
 
     /**
-     * Gets Work week from Severa
-     * 
-     * @param severaUserId Severa user id
-     * @returns Workdays of a user
+     * Fetches workdays for a user from Severa within an optional data range
+     * Modified for calculating user vacation days consumption
```
Jdallos:
```suggestion
```

src/types/severa/workDays/severaResponseWorkDays.ts
```diff
@@ -1,11 +1,15 @@
 /**
  * Interface for Severa Response Work Days.
+ * date and isHoliday fields added so that function can find expected work week
+ * for the logged in user
```
Jdallos:
```suggestion
```

Jdallos (approved):

src/functions/severa/get-filtered-workdays/handler.ts
Alexander-K-byte: That 4 weeks should be enough to guarantee a week without holidays, the tightest pack of holidays occurs at christmas/new year, at max you are looking at 3 weeks of holidays which are Christmas, new year, and epiphany, however I can just extend it to 5 weeks to be safe.

src/functions/severa/get-filtered-workdays/handler.ts
Alexander-K-byte: break added and foundWeek removed since no longer required

Jdallos (approved):

PR #276, opened 2025-11-12
Feature 261 lambdas handler for workdays chart

JuusoHeikkinen02:
Created a new handler that lists severa users workdays for a specific time frame that can be set with startDate and endDate (defaults to week's worth of data if time frame is not set).

Brings the data in a array that has entries consisting of (Date, userGuid,  expectedHours, enteredHours, isHoliday, holidayName).


Alexander-K-byte (approved):

jessesillman (approved):

src/functions/severa/list-workdays-for-user/handler.ts
```diff
@@ -0,0 +1,79 @@
...
+        body: JSON.stringify({ message: "User has not opted in." })
+      };
+    }
+
+    const severaUser = await SeveraApi.getUser(severaUserId);
+
+    const testUserEmail = Config.get().testUser.email;
+    const keycloakEmail = testUserEmail || event.requestContext.authorizer.claims.email;
+
+    if (severaUser.email !== keycloakEmail) {
```
Jdallos, line 42: Consider how we want to handle the admin UI view which will request multiple users. Likely we want a seperate endpoint in which case this is fine.

src/functions/severa/list-workdays-for-user/handler.ts
```diff
@@ -0,0 +1,79 @@
...
+    const startDate = event.queryStringParameters?.startDate;
+    const endDate = event.queryStringParameters?.endDate;
+
+    const rawWorkWeekData: SeveraResponseWorkDays[] = await SeveraApi.getWorkWeek(
+      severaUserId,
+      startDate,
+      endDate
+    );
+
+    const workWeekData = rawWorkWeekData.map((day) => ({
```
Jdallos: We should be able to explicitly type this being returned from teh homelambdas client based on spec

Jdallos (approved):

src/functions/severa/list-workdays-for-user/handler.ts
JuusoHeikkinen02: I will leave this as a reminder

PR #279, opened 2025-11-12
Feature 254 severa opt out

jessesillman:
## What has been done
- Added a remove-opt-in lambda handler.
- Fetches the Keycloak user and, if present, removes the Severa opt-in keyword (`isSeveraOptIn`) and Keycloak attributes (`isSeveraOptIn`, `severaUserId`).
- Added `user:delete` scope to the Severa API access token request (to handle delete operations).


src/database/services/software-service.ts
JuusoHeikkinen02, line 1: Do you know why this has been added?

JuusoHeikkinen02 (approved):
Still need to add the function inside the /src/functions/index.ts file

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    try {
+      await keycloakApi.removeUserAttribute(keycloakUserId, "isSeveraOptIn");
+    } catch {
+      try {
+        await keycloakApi.updateUserAttributes(keycloakUserId, { isSeveraOptIn: [] });
+      } catch {
+        return {
+          statusCode: 502,
```
Alexander-K-byte: Should we be using 5xx responses, those are normally reserved for a broken function or bugged code, perhaps a 400 or 404 would be a better representation?

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+     */
+
+    try {
+      await keycloakApi.removeUserAttribute(keycloakUserId, "severaUserId");
+    } catch {
+      try {
+        await keycloakApi.updateUserAttributes(keycloakUserId, { severaUserId: [] });
+      } catch {
+        return {
+          statusCode: 502,
```
Alexander-K-byte: Same as previous comment

Alexander-K-byte (approved):

src/database/services/software-service.ts
jessesillman: Oh, I never realized that. I think it was added when i merged develop to this branch, though.

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+  try {
+    const severaApi = CreateSeveraApiService();
+    const keycloakApi = CreateKeycloakApiService();
+
+    /**
+     * Fetch Keycloak user and severaUserId attribute
+     */
+    let keycloakUser = await keycloakApi.findUser(keycloakUserId);
+    if (Array.isArray(keycloakUser)) keycloakUser = keycloakUser[0];
+    const attrs = (keycloakUser as any)?.attributes || {};
```
Jdallos: avoid any

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  try {
+    const severaApi = CreateSeveraApiService();
+    const keycloakApi = CreateKeycloakApiService();
+
+    /**
+     * Fetch Keycloak user and severaUserId attribute
+     */
+    let keycloakUser = await keycloakApi.findUser(keycloakUserId);
```
Jdallos: Do we need error handling for this? e.g 
What if no user found

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (Array.isArray(keycloakUser)) keycloakUser = keycloakUser[0];
+    const attrs = (keycloakUser as any)?.attributes || {};
+    const severaUserId: string | undefined = Array.isArray(attrs.severaUserId)
```
Jdallos: How come we don't know if attrs.severaUserId is an array or a string? If we did then we wouldn't need the ternary

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+      ? attrs.severaUserId[0]
+      : attrs.severaUserId;
+
+    /**
+     * If we have Severa user id, try to remove the opt-in keyword from Severa
+     */
```
Jdallos: This comment is fine but should be in // format. The /** format is TSDOcs reserved for functions

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+          };
+        }
+      }
+    }
+
+    try {
+      await keycloakApi.removeUserAttribute(keycloakUserId, "isSeveraOptIn");
+    } catch {
+      try {
+        await keycloakApi.updateUserAttributes(keycloakUserId, { isSeveraOptIn: [] });
```
Jdallos: Why do we try to update if the remove does not work- do you have doubts about this working?

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+          statusCode: 502,
+          body: JSON.stringify({ message: "Failed to remove/clear isSeveraOptIn attribute." })
+        };
+      } 
+    }
+
+    /**
+     * If direct removal fails, fallback by clearing the 'severaUserId' attribute
+     * using updateUserAttributes method to ensure the user is opted out.
+     */
```
Jdallos: // comment not TSDoc and remove the line break after it

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+    } catch {
+      try {
+        await keycloakApi.updateUserAttributes(keycloakUserId, { severaUserId: [] });
+      } catch {
+        return {
+          statusCode: 502,
+          body: JSON.stringify({ message: "Failed to remove severaUserId attribute." }),
+        };
+      }
+    }
```
Jdallos, line 85: Why is this block repeated, Remove if not needed.

src/functions/severa/remove-opt-in/handler.ts
Jdallos: add an appropriate try/ catch

src/services/severa-api-service.ts
```diff
@@ -680,7 +681,32 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+    /**
+     * Removes a keyword from a Severa user.
+     *
+     * @param userGuid The GUID of the user from whom the keyword is being removed.
+     * @param keywordGuid The GUID of the keyword to remove.
+     */
+    removeKeyWordFromUser: async (userGuid: string, keywordGuid: string) => {
+      const removeKeywordUrl = `${baseUrl}/v1/users/${userGuid}/keywords/${keywordGuid}`;
+      const removeResponse = await fetch(removeKeywordUrl, {
+        method: "DELETE",
```
Jdallos, line 695: When we test this we migth need to update the permissions config in severa. Have you tested this locally already?

src/services/severa-api-service.ts
```diff
@@ -698,7 +724,7 @@ const getSeveraAccessToken = async (): Promise<string> => {
     client_id: client_Id,
     client_secret: client_Secret,
     scope:
-      "projects:read, resourceallocations:read, hours:read, users:read, users:write, settings:write, settings:read"
+      "projects:read, resourceallocations:read, hours:read, users:read, users:write, users:delete, settings:write, settings:read"
```
Jdallos, line 727: So probably I do need to update the config, but lets do it after testing.

package-lock.json
Jdallos, line 1: You have no new installs, so any idea why the package-lock diff? I'd suggest running npm i locally just incase you are missing some installs

Jdallos (changes requested):

src/services/severa-api-service.ts
jessesillman: I have tested this locally by running "npx serverless offline --noAuth --stage dev" and calling the endpoint for my user via Postman. Attributes were removed, but toggling the opt-in in the UI, we only receive severaUserID back to the user.

src/functions/severa/remove-opt-in/handler.ts
jessesillman: This block is there to remove the severaUserId attribute if it exists. However, I think we can combine it with the previous block with the severaOptIn to avoid duplication if you prefer.

package-lock.json
jessesillman: Hmm I have no idea, however, I ran the npm i and I no longer see the package-lock diff.

src/functions/severa/remove-opt-in/handler.ts
Jdallos: My mistake, just read it wrong. It's fine

src/services/severa-api-service.ts
Jdallos: That testing sound fine, I would not expect us to be able to test the UI until we have the lambda deployed anyway

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+    };
+  }
+
+  try {
+    const severaApi = CreateSeveraApiService();
+    const keycloakApi = CreateKeycloakApiService();
+
+    /**
+     * Fetch Keycloak user and severaUserId attribute
+     */
```
Jdallos: TSDocs for functions only, this comment (if necessary) should be //

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    const attr = keycloakUser?.attributes || {};
+    severaUserId = attr.severaUserId?.[0];
+
+  } catch {
+    return {
+      statusCode: 502,
+      body: JSON.stringify({ message: "Failed to fetch Keycloak user." }),
+    };
+  }
```
Jdallos: In the github UI the indentation here looks a little off, please check

Jdallos (approved):

PR #280, opened 2025-11-13
Feature 271 get oncall user email

JuusoHeikkinen02:
Created a new GET endpoint to on-call-service folder that fetches the users email by username from the on call API.

The service is called on the weeklycheck handler so it updates all future on call entries with correct email for the user.

Updated the list-on-call handler so it also brings the email from dynamoDB so front end gets correct data.


Alexander-K-byte (approved):

jessesillman (approved):

src/app/config.ts
```diff
@@ -1,25 +1,25 @@
...
+import type { Configuration } from "src/types";
 
 const env = cleanEnv(process.env, {
   AUTH_ISSUER: str(),
   SPLUNK_API_KEY: str(),
   SPLUNK_API_ID: str(),
   SPLUNK_TEAM_ONCALL_URL: str(),
-  SPLUNK_SCHEDULE_POLICY_NAME : str(),
+  SPLUNK_SCHEDULE_POLICY_NAME: str(),
+  SPLUNK_USER_ONCALL_URL: str(),
```
Jdallos: What is this new env for, does it really need to be an env?

Jdallos: If this is just the url from the API docs with no unique information in it then it doesn't need to be an env and should just be used directly in the getUserEmail

Jdallos (approved):

src/app/config.ts
JuusoHeikkinen02: Got it, should only be the url from API Docs. Will remove

PR #291, opened 2025-11-27
Rename severa_base_url, severa_client_id, severa_client_secret to not…

MariaKatariinaH:
… include DEMO. The environment variables are validated in src/app/config.ts. And added to src/types/index.ts.

### DIFF
--- a/src/services/severa-api-service.ts
+++ b/src/services/severa-api-service.ts
@@ -86,7 +86,7 @@ export const CreateSeveraApiService = (): SeveraApiService => {
         method: "GET",
         headers: {
           Authorization: `Bearer ${await getSeveraAccessToken()}`,
-          Client_Id: getSeveraClientId(),
+          Client_Id: process.env.SEVERA_CLIENT_ID,
           "Content-Type": "application/json"
         }
       });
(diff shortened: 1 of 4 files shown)


MariaKatariinaH:
One of the functions doesn't work with the word demo taken off.

src/services/severa-api-service.ts
```diff
@@ -754,6 +753,6 @@ const getSeveraClientId = (): string => {
   if (process.env.STAGE === "production") {
     return process.env.SEVERA_CLIENT_ID;
   } else {
-    return process.env.SEVERA_DEMO_CLIENT_ID;
+    return process.env.SEVERA_CLIENT_ID;
```
Jdallos: So if you see the sonarcube issue it related to this line. What its saying in human friendly language is that this condition is now redundant- the if and else blocks now return the same. So we can remove the condition and just return the env directly.

What would be even better is to remove the function completely (as it doesn't look like it does anything else) and just provide the env directly wherever the function was called previously.

Jdallos (approved):

src/services/severa-api-service.ts
Jdallos: The reason I suggest removing the function as it makes it more clear what is happening eg. giving an env, rather than hiding this behind a layer where we just see a function name getSeveraClientId. Which is a good name and would be fine as well really. If we had 2 or more functions I'd go down the getSeveraClientId function route

PR #297, opened 2025-12-03
feature-296-update-aws-sdk-pacakges-to-same-versions

Alexander-K-byte:
updated all aws-sdk versions to 3.899.0 to avoid possible version conflicts

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,10 @@
     "node": ">=20.0.0"
   },
   "dependencies": {
-    "@aws-sdk/client-dynamodb": "3.882.0",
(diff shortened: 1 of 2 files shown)


package.json
Jdallos, line 1: I don't see why this would only cause an issue for this handler and as such am hesitant to approve.
Are the s3 related libs only used in this handler?
If so what about:
@aws-sdk/client-dynamodb": "3.899.0",
    "@aws-sdk/lib-dynamodb": "3.899.0",


If they are only used in this function then we can approve, otherwise I need more convinving

Alexander-K-byte: Closing branch to investigate issue further to try and find exact cause

PR #301, opened 2025-12-04
Feature 285 refactor opt in endpoints

jessesillman:
## What has been done
- PUT /users/{userId}/severa-opt-in handler added in home-lambdas.
-  Resolves Severa user by their Keycloak id.
- Ensures that _severaUserId_ is set back to Keycloak attributes, and _isSeveraOptIn_ for the Severa client.


src/functions/severa/add-opt-in/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+    const keyword = await severaApi.checkKeywordExists("isSeveraOptIn");
+    const keywordGuid = keyword && (keyword.guid);
+    if (!keywordGuid) {
+      return { 
+        statusCode: 502, 
+          body: JSON.stringify({ message: "Failed to add the isSeveraOptIn keyword." }) 
+        };
+      }
+
+    await severaApi.updateSeveraOptInKeyword(severaUserId, "true", keywordGuid);
```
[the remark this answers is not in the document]
jessesillman: Leave this open because we might want to keep using updateSeveraOptInKeyword method over addKeyWordToUser that I had previously created for testing.

src/functions/severa/add-opt-in/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+    // Find Severa user by Keycloak id
+    if (!severaUserId) {
+      try {
+        const severaUser = await severaApi.fetchUserByKeycloakId(keycloakUserId);
+        severaUserId = severaUser && (severaUser.guid);
+      } catch {
+        return { 
+          statusCode: 502, 
+          body: JSON.stringify({ message: "Failed to fetch Severa user by Keycloak id." }) 
+        };
```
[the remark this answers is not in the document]
jessesillman: Leave this open. We might want to have these simplified error block, and we use console logs for troubleshooting during development processes.

src/functions/severa/add-opt-in/handler.ts
```diff
@@ -0,0 +1,66 @@
...
+    await severaApi.updateSeveraOptInKeyword(severaUserId, "true", keywordGuid);
+
+    return { 
+      statusCode: 200, 
+      body: JSON.stringify({ message: "Opt-in completed." }) 
+    };
+  } catch {
+    return { 
+      statusCode: 500, 
+      body: JSON.stringify({ message: "Opt-in failed." })
```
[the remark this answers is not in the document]
jessesillman: Same with this one; we prefer having generic error messages for production.

Alexander-K-byte (approved):

Dshad-Anon (approved):

JuusoHeikkinen02 (approved):

src/functions/severa/add-opt-in/handler.ts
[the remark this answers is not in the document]
Jdallos: This suggestion seems reasonable to me, we can keep the generic error message but we could also pass any actual error.
Console.error should be removed

src/services/severa-api-service.ts
```diff
@@ -705,7 +706,38 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+    },
+
+    /**
+     * Fetches a Severa user by their Keycloak id.
+     *
+     * @param keycloakId The Keycloak id of the user to be fetched.
+     * @returns The Severa user object.
+     */
+    fetchUserByKeycloakId: async (keycloakId: string) => {
+      const url = `${baseUrl}/v1/users?keycloakId=${encodeURIComponent(keycloakId)}`;
```
Jdallos: I'm pretty certain that the severaAPI (the base url you are using here) does not fetch users by keycloak Id. So I guess this will just 404?

src/functions/severa/add-opt-in/handler.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  try {
+    const severaApi = CreateSeveraApiService();
+    const keycloakApi = CreateKeycloakApiService();
+
+    // Get severaUserId from Keycloak
+    let severaUserId: string | undefined;
+    try {
+      const severaUser = await severaApi.fetchUserByKeycloakId(keycloakUserId);
```
Jdallos: See my comment for this method i the severa service, but basically pretty sure this won't work.

I would guess the flow for adding opt in is as follows:
-Find the associated Severa user (the the keycloak user making this request), previously this was done via email I believe. I don't think severa will have a reference to keycloak user Id. There might be other ways to identify a non opted in users severa user than email- but I don't know it and likely not keycloakId
-Add the keyword to that severa user
-Get the severaId for the severa user you found and add it to the keycloak user

src/functions/severa/add-opt-in/handler.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    await keycloakApi.updateUserAttributes(keycloakUserId, { severaUserId: [severaUserId] });
+
+    // Add the isSeveraOptIn keyword to Severa user
+    const keyword = await severaApi.checkKeywordExists("isSeveraOptIn");
+    const keywordGuid = keyword.guid;
+    if (!keywordGuid) {
+      return { 
+        statusCode: 502, 
+          body: JSON.stringify({ message: "Failed to add the isSeveraOptIn keyword." })
```
Jdallos: More specific and useful error message here would be Failed to find the severa keyword

src/functions/severa/add-opt-in/handler.ts
```diff
@@ -0,0 +1,62 @@
...
+import { CreateSeveraApiService } from "src/services/severa-api-service";
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import { middyfy } from "src/libs/lambda";
+
+/**
+ * Lambda handler for adding a user's Severa opt-in (keyword and Keycloak attribute).
+ *
+ * @param event API Gateway event containing the userId path parameter.
+ */
+export const addSeveraOptInHandler: APIGatewayProxyHandler = async (event) => {
```
Jdallos, line 11: This function is to replace the updateUserAttribute handler correct? If so that function should also be removed as part of this PR.

src/functions/severa/add-opt-in/handler.ts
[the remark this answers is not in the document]
Jdallos: The Ai is confusing here, I don't see any addKeywordToUser method. However it could be appropriate to rename the updateSeveraOptInKeyword to addKeywordToUser

src/services/severa-api-service.ts
Jdallos: So correct me if I'm wrong but I don't see how this function can work, as such won't comment on the rest of this function.

Jdallos (changes requested):

src/functions/severa/add-opt-in/handler.ts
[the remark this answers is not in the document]
jessesillman: I had implemented that method, but decided to remove and use the initial method that was already implemented. However, I can rename it to addKeywordToUser.

src/services/severa-api-service.ts
jessesillman: You are absolutely right. This should definitely be removed.

src/functions/severa/add-opt-in/handler.ts
jessesillman: You are most likely right, and I will look into this part more carefully, and figure how to implement this part. This required many hours of trying to successfully fetch the severaUserId and isSeveraOptIn keyword to my user, and this is how managed to do it. I will check all the operations and find the logic that resolves this issue.

src/functions/severa/add-opt-in/handler.ts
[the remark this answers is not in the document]
Jdallos: Seems reasonable to me to add the console error as suggested

src/functions/severa/add-opt-in/handler.ts
```diff
@@ -0,0 +1,119 @@
...
+      body: JSON.stringify({ error: "Keycloak userId is required" }) 
+    };
+  }
+
+  try {
+    const severaApi = CreateSeveraApiService();
+    const keycloakApi = CreateKeycloakApiService();
+
+    // Fetch Keycloak user and email
+    let chosenEmail: string | undefined;
```
Jdallos: Name could be improved e.g. userEmail or keycloakUserEmail?

src/services/severa-api-service.ts
```diff
@@ -705,6 +707,58 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
       }
+    },
+
+    /**
+     * Fetches a Severa user by their Keycloak id.
+     *
+     * @param keycloakId The Keycloak id of the user to be fetched.
+     * @returns The Severa user object.
+     */
+    fetchUserByKeycloakId: async (keycloakId: string) => {
```
Jdallos: So this function is wrong, as discussed previously severaApi does not have the functinoanlity to fetch a user by the keycloak Id. At least in this Pr I don't see this function being used and so it should be removed completely.

src/services/severa-api-service.ts
```diff
@@ -705,6 +707,58 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
+
+      const users = await response.json();
+      if (!users?.length) {
+        throw new Error(`No user found with keycloakId: ${keycloakId}`);
+      }
+
+      return users[0];
+    },
+
+    addKeywordToUser: async (userGuid: string, keywordGuid: string) => {
```
Jdallos, line 717: Add TSDocs

src/utils/severa/index.ts
```diff
@@ -10,12 +11,11 @@ import {CreateSeveraApiService} from "src/services/severa-api-service";
  * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
  */
 export const optInSeveraUser = async (email: string, keyword: Record<string, string[]>) => {
-
-  const userEmail = Config.get().testUser.email || email;
+  const userEmail = Config.get().testUser.email;
```
Jdallos: I'm a bit confused by the use of the env userEmail and the argument email-  isn't it that if the env is present then we should use that, in which case I'm not sure why you have made these changes as now you always fetchUserByEmail from the param email (line 18) but always return the env email (line 40).

Shouldn't this be how it was so :
const userEmail = Config.get().testUser.email || email;

And then use userEmail on both line 18 and 40?
Let me know if you disagree

Jdallos (approved):

src/utils/severa/index.ts
jessesillman: Switched it back to how it was. Probably just forgot to discard changes for that when I tried testing/reconfiguring.

PR #302, opened 2025-12-04
feature-296-more-bug-checking-for-upload-image

Alexander-K-byte:
added check for bucket name / region to ensure they are not the cause of the issue

### DIFF
--- a/src/services/s3-file-service.ts
+++ b/src/services/s3-file-service.ts
@@ -1,7 +1,17 @@
 import { GetObjectCommand, PutObjectCommand, S3Client } from "@aws-sdk/client-s3";
 import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
 
-const { HOME_BUCKET_NAME, HOME_BUCKET_REGION } = process.env;
+const HOME_BUCKET_NAME = process.env.HOME_BUCKET_NAME;
+const HOME_BUCKET_REGION = process.env.HOME_BUCKET_REGION;
+
+if (!HOME_BUCKET_NAME) {
+  throw new Error("HOME_BUCKET_NAME is not defined in environment variables");
+}
+
+if (!HOME_BUCKET_REGION) {
+  throw new Error("HOME_BUCKET_REGION is not defined in environment variables");
+}
+
 const s3Client = new S3Client({ region: HOME_BUCKET_REGION });
 
 /**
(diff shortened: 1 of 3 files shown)


MariaKatariinaH (approved):

JuusoHeikkinen02 (approved):

src/services/s3-file-service.ts
```diff
@@ -1,16 +1,30 @@
...
+const HOME_BUCKET_NAME = process.env.HOME_BUCKET_NAME;
+const HOME_BUCKET_REGION = process.env.HOME_BUCKET_REGION;
+
+if (!HOME_BUCKET_NAME) {
+  throw new Error("HOME_BUCKET_NAME is not defined in environment variables");
+}
+
+if (!HOME_BUCKET_REGION) {
+  throw new Error("HOME_BUCKET_REGION is not defined in environment variables");
+}
```
Jdallos: This is unnecessary as we already have env validation se t up in the app/config, these envs are missing from there howeever so should be added in there

src/services/s3-file-service.ts
```diff
@@ -1,16 +1,30 @@
...
+
+const s3Client = new S3Client({
+  region: HOME_BUCKET_REGION,
+  forcePathStyle: false,
+  defaultsMode: "standard"
```
Jdallos: Looks like you are changing this and the checksumAlgorithy in the PutObjectCommand. Why are you making these changes?
I struggled to find clear references to these- I'm not sure forcePathStyle is on v3, and defaultsMode standard is the default option?

Jdallos (changes requested):

src/services/s3-file-service.ts
Alexander-K-byte: The documentation does not provide any information for this, searches online have people saying this was required, another said they just had to set the algorithm as undefined in the PutObjectCommand, others have said that they had to use Checksum Disabled. 

As previously mentioned, this is at best a patch attempt to disable Checksum which is a best guess for the issue currently. And  would probably require implementation of checksum calculations frontend which can be sent to be compared in the backend if this is indeed the issue, either that or we use this bucket with checksums disabled.

src/services/s3-file-service.ts
Alexander-K-byte: Suggested change made

Jdallos (approved):

PR #303, opened 2025-12-08
Bugfix-keycloak-vacation-days-consumption-incorrect

Alexander-K-byte:
Directly map 6 day logic to daysByYear and also assumes vacation year runs from 1st April -> 31st March.

### DIFF
--- a/src/libs/vacation-utils.ts
+++ b/src/libs/vacation-utils.ts
@@ -71,24 +71,17 @@ export const splitVacationDaysByYear = (
   // Calculate weeks & days of request for 6 day work week logic
   const fullWeeks = Math.floor(workDaysInRange / workDays);
   const extraDays = workDaysInRange % workDays;
-
   const totalDays = fullWeeks * 6 + extraDays;
 
+  // Assumes vacation year runs 01/04/YYYY → 31/03/(YYYY+1)
+  const getVacationYear = (date: DateTime): string => {
+    return date.month >= 4 ? String(date.year) : String(date.year - 1);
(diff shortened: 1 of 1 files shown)


JuusoHeikkinen02 (commented):
If the issue was around calculating the correct days (including saturdays when needed). Does this fix it yet or is this just to check that the function consumes for correct year?  (edit) Looks like it does

JuusoHeikkinen02 (approved):

Dshad-Anon (approved):

src/libs/vacation-utils.ts
```diff
@@ -71,24 +71,17 @@ export const splitVacationDaysByYear = (
...
   const totalDays = fullWeeks * 6 + extraDays;
 
-  const daysByYear: Record<string, number> = {};
-  currentDate = startDateObj;
-  let daysLeft = totalDays;
-
-  while (currentDate <= endDateObj && daysLeft > 0) {
-    const year = currentDate.year.toString();
-    if (!daysByYear[year]) daysByYear[year] = 0;
+  // Assumes vacation year runs 01/04/YYYY → 31/03/(YYYY+1)
```
Jdallos, line 76: Why are we assuming this? Are you sure its correct, I think they run january to january as I said previously

Alexander-K-byte: All sources available online say that the holiday credit year is from 1st April till 31st March, unless this is separate to holiday year?

Jdallos (approved):

PR #304, opened 2025-12-10
Feature 288 memos translation to dynamodb

JuusoHeikkinen02:
### Restore
Restored management memo endpoints for listing/fetching content from the google drive.

### Feature
Created a service for getting personal google authentication for accessing the drive (Service account will be implemented later so this will probably be removed).

Created a TranslatedMemoService for dynamoDB that has storeMemo and getTranslatedPdf functions.

Created handlers for creating/storing the translated memos and getting the translated memos from dynamoDB by id and targetLanguage.

### TODO
The translation implementation is not ready yet because we don't have the service for translation. There is commented code in the creation handler for the translation and I have tested it with mock data to ensure the sorting by targetLanguage works.


src/services/google-drive-api-service.ts
```diff
@@ -210,60 +247,70 @@ export const getFolderId = async (year: string, month: string): Promise<string>
...
-    )
+    const response = await drive.files.get(
+      {
+        fileId: file.id,
+        alt: "media"
+      },
+      { responseType: "stream" }
+    );
     if (!response) {
       console.error(`Failed to fetch file content: ${response.status} - ${response.statusText}`);
```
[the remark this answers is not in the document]
JuusoHeikkinen02: removed unnecessary error catches, try catch should be fine for catching the errors

src/services/google-drive-api-service.ts
```diff
@@ -102,12 +124,12 @@ const getFolders = async (folderId: string) => {
 export const getBaseFolderByName = async (folderName: string) => {
   try {
     const folders = await getFolders(folderId);
-    const folderFound = folders.find(folder => folder.name === folderName);
+    const folderFound = folders.find((folder) => folder.name === folderName);
     return folderFound.id;
```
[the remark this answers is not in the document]
JuusoHeikkinen02: Added a descriptive error

Alexander-K-byte (approved):

MariaKatariinaH (approved):

jessesillman (approved):

src/app/config.ts
```diff
@@ -47,6 +51,12 @@ export default class Config {
       mailgunSmtpPassword: env.MAILGUN_SMTP_PASSWORD,
       adminEmails: env.ADMIN_EMAILS ? env.ADMIN_EMAILS.split(",") : []
     },
+    google: {
+      clientId: env.GOOGLE_CLIENT_ID,
+      clientSecret: env.GOOGLE_CLIENT_SECRET,
+      refreshToken: env.GOOGLE_REFRESH_TOKEN,
+      driveFolderId: env.GOOGLE_DRIVE_FOLDER_ID
```
Jdallos, line 61: Are these likely to need changing when we get access to the workplace API? If so we should leave a comment

src/database/models/memo-record.ts
```diff
@@ -0,0 +1,20 @@
...
+  id: string;
+  PK: string;
+  SK: string;
+  fileId: string;
+  fileName: string;
+  language: string;
+  translatedBase64: string;
+}
+
+export interface MemoInput {
```
Jdallos, line 17: Missing comment

src/database/services/translated-memo-api-service.ts
```diff
@@ -0,0 +1,60 @@
+import { type DynamoDBDocumentClient, GetCommand, PutCommand } from "@aws-sdk/lib-dynamodb";
+import type { MemoInput, MemoRecord } from "src/database/models/memo-record";
+import { v4 as uuidv4 } from "uuid";
+
+const TABLE_NAME = "TranslatedMemos";
+
+class TranslatedMemoService {
```
Jdallos, line 12: Comment?

src/database/services/translated-memo-api-service.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Retrieves a translated PDF memo from DynamoDB.
+   *
+   * @param id - The primary key of the memo (`MEMO#id`).
+   * @param targetLanguage - The target language of the memo (`LANG#language`).
+   *
+   * @returns The `MemoRecord` if found, or `null` if no matching memo exists.
+   */
+  public getTranslatedPdf = async (
```
Jdallos: Is it actually a Pdf we retrieve?

src/functions/memo-management/drive-memos/translated-memos/create-translated-memo-pdf/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "POST",
+        path: "/translate-memo",
```
Jdallos, line 9: So this creates the translated memo, and expects to be provided a file id, but as this is a lambda function we are expecting to call it from the UI? So how will that work in terms of how does the user get the fileId (the google drive file) in the first place?
This might require some additional planning/ consulting the original thesis notes- do we list the files directly from the drive in the UI?

src/services/google-drive-api-service.ts
```diff
@@ -20,59 +21,78 @@ export const getGoogleAuth = async () => {
...
   }
-}
+};
+
+/**
+ * Personal Google authentication service (!!for testing the drive API!!)
+ *
+ * @returns Google Authentication
+ */
+export const getPersonalGoogleAuth = async () => {
```
Jdallos, line 37: So this will be redundant when we move to workspace API? IF so might be worth a comment

src/services/google-drive-api-service.ts
```diff
@@ -116,27 +143,35 @@ export const getBaseFolderByName = async (folderName: string) => {
  * @param mimeType type of file, PDF format by default
  * @returns array of File objects
  */
-export const getFilesInYear = async (year: string, mimeType: string = "application/pdf"): Promise<File[]> => {
+export const getFilesInYear = async (
```
Jdallos, line 147: So obviously the structure of these functions imply we expect some strucutre to the google drive? If so it could be worth documenting this- perhaps in Alex's onboarding docs but also in the api spec in the related endpoint

src/services/google-drive-api-service.ts
```diff
@@ -210,60 +252,66 @@ export const getFolderId = async (year: string, month: string): Promise<string>
...
   } catch (error) {
     console.error("Error loading the PDF:", error);
   }
-}
+};
 
 /**
  * Creates a new PDF file from binary content and stores it in a specified folder
- * 
+ *
```
Jdallos: We should probably remove this createPdfFile and all references as I'm pretty sure we don't want to be creating anything in the drive- read only

Jdallos (approved):

src/database/services/translated-memo-api-service.ts
JuusoHeikkinen02: Well in this case we are retrieving the whole memorecord so might need to change the services name accordingly, also will make some changes to the handler so that we are sending the data in a actual pdf stream and not as JSON that contains the encoded base64 of the pdf.

src/functions/memo-management/drive-memos/translated-memos/create-translated-memo-pdf/index.ts
JuusoHeikkinen02: Yes the folders and files are listed first on the UI by year. Then the services fetch all folders/files by that year param so all of the needed fileIds will be accessible for the handler. One thing we need to consider is that when do we want to call this function exactly: when user selects the pdf from the drive to be displayed or only after user clicks the translation button on that pdf's page?

PR #311, opened 2025-12-17
Restoring handler for getting users slack avatar

JuusoHeikkinen02:
Restored/modified the existing handler for getting users slack avatar. Previously the handler was using a service that fetched all users form the slack workspace and it had retry/rate limit helpers but those are not needed anymore because its now only fetching the image for the loggedIn user.

(lot of formatting on the slack utils file, only made changes to the getSlackUserByEmail and added the user override and channelId envs to config)


Alexander-K-byte (approved):

jessesillman (approved):

src/functions/slack/get-slack-user-avatar/handler.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Lambda to receive user's slack avatar by email
+ *
+ * @param event event containing path parameter 'email'
+ * @return A response object with statuscode and avatar URL
+ */
+const getSlackUserAvatarHandler: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {
+  try {
+    const { accessToken } = await Auth.getAccessToken();
```
Jdallos: Isn't this typically accessed via the event in other lambdas?

src/functions/slack/get-slack-user-avatar/index.ts
```diff
@@ -0,0 +1,16 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+export default {
+  handler: `${handlerPath(__dirname)}/handler.main`,
+  events: [
+    {
+      httpApi: {
+        method: "get",
+        path: "/slack-avatar/{email}",
```
Jdallos: Seems a bit odd to have the email in the path, I think typical REST convention would be to have it as a query param, might be worth checking best REST practice.

As email is basically the look up key then I strongly recommend using it as query param

src/meta-assistant/slack/slack-utils.ts
```diff
@@ -207,23 +234,22 @@ namespace SlackUtilities {
...
+        messageResults.push({
+          message: message,
+          response: await sendMessage(slackId, message.message)
+        });
+      } else {
+        for (const stagingid of slackOverride) {
           messageResults.push({
             message: message,
-            response: await sendMessage(slackId, message.message)
+            response: await sendMessage(stagingid, message.message)
```
Jdallos: Are you sure about this change, as far as I can tell it's not related to this task- have you actually tested this works?

src/meta-assistant/slack/slack-utils.ts
```diff
@@ -266,16 +296,19 @@ namespace SlackUtilities {
...
+  ): Promise<NotificationMessageResult> => {
     let message;
-      message = await constructMemoDocCreatedMessage(summary, name);
-  
+    message = await constructMemoDocCreatedMessage(summary, name);
+
     const messageResults: NotificationMessageResult = {
       message: message,
-      response: await sendMessage(slackChannelId, message)
+      response: await sendMessage(channelId, message)
```
Jdallos, line 308: And again are you sure about this change as the function is not related to the task this feature addresses.

Jdallos (changes requested):

src/meta-assistant/slack/slack-utils.ts
JuusoHeikkinen02: I'm really not sure why this has changed to stagingid, I don't think the function was working with the slackId. I will still revert this change.

src/meta-assistant/slack/slack-utils.ts
JuusoHeikkinen02: For this I changed it to use the configs cleanenv variable to be consistent with other implementations and it still should work because its the same env.  Can change it back to the original implementation if this is not appropriate with my current task.

src/meta-assistant/slack/slack-utils.ts
Jdallos: Could be good further work to investigate the MetaAssistant related functions and make sure they work (this is part of that) something to keep in mind if you run out of tasks maybe

Jdallos (approved):

PR #317, opened 2026-01-07
bugfix-316-prevent-duplicate-memo-creation

JuusoHeikkinen02:
Prevent duplicate memo creation by adding a Global Secondary Index and a retrieval method for existing memos by file ID and language.

### DIFF
--- a/serverless.ts
+++ b/serverless.ts
@@ -329,11 +329,27 @@ const serverlessConfiguration: AWS = {
           TableName: "Memos",
           AttributeDefinitions: [
             { AttributeName: "PK", AttributeType: "S" },
-            { AttributeName: "SK", AttributeType: "S" }
+            { AttributeName: "SK", AttributeType: "S" },
+            { AttributeName: "fileId", AttributeType: "S" },
+            { AttributeName: "language", AttributeType: "S" }
           ],
           KeySchema: [
             { AttributeName: "PK", KeyType: "HASH" },
             { AttributeName: "SK", KeyType: "RANGE" }
+          ],
+          GlobalSecondaryIndexes: [
+            {
+              IndexName: "FileIdIndex",
+              KeySchema: [
+                { AttributeName: "fileId", KeyType: "HASH" },
+                { AttributeName: "language", KeyType: "RANGE" }
+              ],
+              Projection: { ProjectionType: "ALL" },
+              ProvisionedThroughput: {
+                ReadCapacityUnits: 1,
+                WriteCapacityUnits: 1
+              }
+            }
           ],
           ProvisionedThroughput: {
             ReadCapacityUnits: 1,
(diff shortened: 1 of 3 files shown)


src/functions/memo-management/drive-memos/translated-memos/create-translated-memo-pdf/handler.ts
```diff
@@ -41,51 +40,54 @@ const createTranslatedMemoPdfHandler: APIGatewayProxyHandlerV2 = async (
...
     // const translatedBase64 = translatedPdf.content.toString("base64");
-    //   const translatedBase64 = "is this translated?";
-    //   const translatedMemo = {
-    //     id: storedMemo.id,
-    //     fileId: file.id,
-    //     fileName: file.name,
-    //     language: "en", // or dynamically from translation service
-    //     translatedBase64: translatedBase64
-    //   };
+    // const translatedBase64 = "is this not translated?";
```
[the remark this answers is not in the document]
JuusoHeikkinen02: Not really important, because that check will be implemented with actual data rather than with this comment

MariaKatariinaH (approved):

Dshad-Anon (approved):

serverless.ts
```diff
@@ -156,7 +156,8 @@ const serverlessConfiguration: AWS = {
                   "arn:aws:dynamodb:${self:provider.region}:*:table/Articles",
                   "arn:aws:dynamodb:${self:provider.region}:*:table/Articles/index/GSI_Path",
                   "arn:aws:dynamodb:${self:provider.region}:*:table/OnCallSchedule",
-                  "arn:aws:dynamodb:${self:provider.region}:*:table/Memos"
+                  "arn:aws:dynamodb:${self:provider.region}:*:table/Memos",
+                  "arn:aws:dynamodb:${self:provider.region}:*:table/Memos/index/FileIdIndex"
```
Jdallos, line 160: Do we really need this? And similar question for the :
 "arn:aws:dynamodb:${self:provider.region}:*:table/Articles/index/GSI_Path",

Please check

Jdallos (approved):

serverless.ts
JuusoHeikkinen02: Yes in this case its needed because the translation handler is now doing a query to the index and those arn's don't inherit permissions. There is a another way for giving permissions but it would be like:
"arn:aws:dynamodb:${self:provider.region}:*:table/Memos*"

then it would cover the table and all GSIs if in the future there would be a need for more GSIs

Jdallos: Lets leave as it is and try it for now, if we run into issues lets then go wtih:
"arn:aws:dynamodb:${self:provider.region}::table/Memos"

PR #321, opened 2026-01-14
Ensure the validation check is there before database updates.

### LINKED ISSUE #320: Database Updated Before vacation Days validation
The vacation request was being persisted to database before validating whether the user has sufficient vacation days. If validation fails after database update, then vacation request would be on Approved state in database but users vacation days in key cloak will not be deducted creating data inconsistency. 

Also when a user edits already approved vacation request it never restores /add back the vacation days of the OLD dates. 
So when a vacation request is rejected or cancelled after being approved, vacation days should be restored to the user.

Dshad-Anon:
Validation works for this. The user cannot create requests with more days than available.

### DIFF
--- a/src/functions/vacation-request/create-vacation-request/handler.ts
+++ b/src/functions/vacation-request/create-vacation-request/handler.ts
@@ -75,7 +75,7 @@ export const createVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
       const hasEnoughDays = await validateVacationDays(userId, daysInYear, year);
       if (!hasEnoughDays) {
         return {
-          statusCode: 400,
+          statusCode: 409,
           body: JSON.stringify({
             message: `Cannot create request: you do not have enough remaining vacation days for ${year}.`
           })
--- a/src/functions/vacation-request/create-vacation-request/handler.ts
+++ b/src/functions/vacation-request/create-vacation-request/handler.ts
@@ -5,7 +5,7 @@ import { middyfy } from "src/libs/lambda";
 import {
   getContractedWeek,
   splitVacationDaysByYear,
-  updateRemainingVacationDays
+  validateVacationDays
 } from "src/libs/vacation-utils";
 import { notifyAdminsVacationSubmittedAll } from "src/notifications/vacation-notifications";
 import type vacationRequestSchema from "src/schema/vacationRequest";
@@ -68,6 +68,21 @@ export const createVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
   const userDetails = await api.findUser(userId);
 
   try {
+    const contractedWeek = await getContractedWeek(userId);
+    const daysByYear = splitVacationDaysByYear(startDate, endDate, contractedWeek);
+
+    for (const [year, daysInYear] of Object.entries(daysByYear)) {
+      const hasEnoughDays = await validateVacationDays(userId, daysInYear, year);
+      if (!hasEnoughDays) {
+        return {
+          statusCode: 400,
+          body: JSON.stringify({
+            message: `Cannot create request: you do not have enough remaining vacation days for ${year}.`
+          })
+        };
+      }
+    }
+
     const createdVacationRequest = await vacationRequestService.createVacationRequest({
       id: newVacationRequestId,
       userId: userId,
@@ -82,33 +97,6 @@ export const createVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
       createdAt: createdAt,
       updatedAt: updatedAt
     });
-
-    const contractedWeek = await getContractedWeek(userId);
-    const daysByYear = splitVacationDaysByYear(startDate, endDate, contractedWeek);
-    const currentStatus = Array.isArray(status) ? status.at(-1)?.status : "UNKNOWN";
-
-    try {
-      for (const [year, daysInYear] of Object.entries(daysByYear)) {
-        const success = await updateRemainingVacationDays(userId, daysInYear, currentStatus, year);
-
-        if (!success) {
-          return {
-            statusCode: 400,
-            body: JSON.stringify({
-              message: `Cannot create request: user does not have enough remaining vacation days for ${year}.`
-            })
-          };
-        }
-      }
-    } catch (err) {
-      console.error("Vacation update failure:", err);
-      return {
-        statusCode: 500,
-        body: JSON.stringify({
-          message: "Unexpected error while updating remaining vacation days."
-        })
-      };
-    }
 
     if (draft === false) {
       await notifyAdminsVacationSubmittedAll({
--- a/src/functions/vacation-request/update-vacation-request/handler.ts
+++ b/src/functions/vacation-request/update-vacation-request/handler.ts
@@ -120,6 +120,11 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
   };
 
   try {
+    /**
+     * Handles vacation request approval flow.
+     * Deducts vacation days from user's balance before updating the database.
+     * Implements compensating transaction: if database update fails, vacation days are returned.
+     */
     if (currentStatus === VacationRequestStatuses.Approved && statusChanged) {
       const validationError = await validateVacationApproval(
         userId,
@@ -129,18 +134,30 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
       );
       if (validationError) return validationError;
 
-      const updatedVacationRequest =
-        await vacationRequestService.updateVacationRequest(vacationRequestUpdates);
-
       await deductVacationDaysForApproval(userId, startDate, endDate, contractedWeek);
-      await sendNotifications();
 
-      return {
-        statusCode: 200,
-        body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
-      };
+      try {
+        const updatedVacationRequest =
+          await vacationRequestService.updateVacationRequest(vacationRequestUpdates);
+
+        await sendNotifications();
+
+        return {
+          statusCode: 200,
+          body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
+        };
+      } catch (dbError) {
+        console.error("Database update failed, rolling back vacation day deduction", dbError);
+        await returnVacationDaysForRejection(userId, startDate, endDate, contractedWeek);
+        throw dbError;
+      }
     }
 
+    /**
+     * Handles vacation request rejection/cancellation flow.
+     * Updates database status before returning vacation days to user's balance.
+     * If returning days fails, logs error but considers operation successful as status change takes priority.
+     */
     if (
       existingLatestStatus === VacationRequestStatuses.Approved &&
       statusChanged &&
@@ -149,18 +166,28 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
       const updatedVacationRequest =
         await vacationRequestService.updateVacationRequest(vacationRequestUpdates);
 
-      await returnVacationDaysForRejection(
-        userId,
-        existingVacationRequest.startDate,
-        existingVacationRequest.endDate,
-        contractedWeek
-      );
-      await sendNotifications();
+      try {
+        await returnVacationDaysForRejection(
+          userId,
+          existingVacationRequest.startDate,
+          existingVacationRequest.endDate,
+          contractedWeek
+        );
 
-      return {
-        statusCode: 200,
-        body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
-      };
+        await sendNotifications();
+
+        return {
+          statusCode: 200,
+          body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
+        };
+      } catch (keycloakError) {
+        console.error("Failed to return vacation days after status change.", keycloakError);
+
+        return {
+          statusCode: 200,
+          body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
+        };
+      }
     }
 
     const updatedVacationRequest =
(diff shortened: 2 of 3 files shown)


src/libs/vacation-utils.ts
```diff
@@ -88,32 +88,47 @@ export const splitVacationDaysByYear = (
...
- * @returns A promise that resolves when the user's vacation days have been updated.
+ * @returns True if user has enough days, false otherwise.
  */
-export const updateRemainingVacationDays = async (
+export const validateVacationDays = async (
   userId: string,
-  daysToSubtract: number,
-  status: string,
+  daysNeeded: number,
   year: string
```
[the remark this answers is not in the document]
Dshad-Anon: Its not allowed in frontend to pass negative value. That will ensure its always positive number.

src/libs/vacation-utils.ts
```diff
@@ -88,32 +88,47 @@ export const splitVacationDaysByYear = (
...
+/**
+ * Helper function to deductVacationDays - modifies Keycloak to deduct vacation days.
+ * @param userId - The ID of the user whose vacation days are being deducted.
+ * @param daysToDeduct - The number of vacation days to deduct.
+ * @param year - The year for which the vacation days are being deducted.
+ */
+export const deductVacationDays = async (
+  userId: string,
+  daysToDeduct: number,
+  year: string
```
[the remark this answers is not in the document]
Dshad-Anon: Always postive number is allowed to be requested from frontend so leaving it.

jessesillman (approved):

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -126,6 +117,57 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
...
+      currentStatus !== VacationRequestStatuses.Approved
+    ) {
+      const updatedVacationRequest =
+        await vacationRequestService.updateVacationRequest(vacationRequestUpdates);
+
+      await returnVacationDaysForRejection(
+        userId,
+        existingVacationRequest.startDate,
+        existingVacationRequest.endDate,
+        contractedWeek
```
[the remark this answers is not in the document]
Dshad-Anon: user cannot change the date and status at the same time in the frontend which resolves the speculation of making inconsitency.

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -126,6 +117,57 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
...
+      );
```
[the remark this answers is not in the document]
Dshad-Anon: The new logic should look after this inconsitency.

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -126,6 +117,84 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
...
       });
     }
+  };
+
+  try {
+    /**
+     * Handles vacation request approval flow.
+     * Deducts vacation days from user's balance before updating the database.
+     * Implements compensating transaction: if database update fails, vacation days are returned.
+     */
```
JuusoHeikkinen02: remove tsdoc from here if you want to comment the workflow use //

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -126,6 +117,84 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
...
+        await returnVacationDaysForRejection(userId, startDate, endDate, contractedWeek);
+        throw dbError;
+      }
+    }
+
+    /**
+     * Handles vacation request rejection/cancellation flow.
+     * Updates database status before returning vacation days to user's balance.
+     * If returning days fails, logs error but considers operation successful as status change takes priority.
+     */
```
JuusoHeikkinen02: I guess same here

JuusoHeikkinen02 (approved):

src/functions/vacation-request/update-vacation-request/handler.ts
Dshad-Anon: new commit makes the new function and use of tsdocs

src/functions/vacation-request/create-vacation-request/handler.ts
```diff
@@ -68,6 +68,21 @@ export const createVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
...
 
   try {
+    const contractedWeek = await getContractedWeek(userId);
+    const daysByYear = splitVacationDaysByYear(startDate, endDate, contractedWeek);
+
+    for (const [year, daysInYear] of Object.entries(daysByYear)) {
+      const hasEnoughDays = await validateVacationDays(userId, daysInYear, year);
+      if (!hasEnoughDays) {
+        return {
+          statusCode: 400,
```
Jdallos: 409 might be more appropriate, up to you

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
 } from "src/libs/vacation-utils";
 import {
   notifyAdminsVacationSubmittedAll,
   notifyUserVacationStatusUpdatedAll
 } from "src/notifications/vacation-notifications";
 import type vacationRequestSchema from "src/schema/vacationRequest";
 
+/**
+ * Handles vacation request approval with compensating transaction.
+ */
```
Jdallos, line 29: Include param docs

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
+const handleApproval = async (
```
Jdallos, line 30: HandleApproval and handleRejection should be moved into a utils (utils/vacations?) file so we keep this handler file free for just the handler. Unless there is some strong reason for them to be here?

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
+  validateVacationApproval
 } from "src/libs/vacation-utils";
 import {
   notifyAdminsVacationSubmittedAll,
   notifyUserVacationStatusUpdatedAll
 } from "src/notifications/vacation-notifications";
 import type vacationRequestSchema from "src/schema/vacationRequest";
 
+/**
+ * Handles vacation request approval with compensating transaction.
```
Jdallos: This description could be improved. e.g. verifies user has sufficient vacation days and deducts them

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
+  await deductVacationDaysForApproval(userId, startDate, endDate, contractedWeek);
+
+  try {
+    const updatedVacationRequest =
+      await vacationRequestService.updateVacationRequest(vacationRequestUpdates);
+    await sendNotifications();
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
```
Jdallos, line 55: Can this casting as unknown be done in a better way, why is there such a mismatch between the types that this is needed? Could be further work

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
+    };
+  } catch (dbError) {
```
Jdallos, line 57: typically just error for the param name here is best

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
+    console.error("Database update failed, rolling back vacation day deduction", dbError);
+    await returnVacationDaysForRejection(userId, startDate, endDate, contractedWeek);
+    throw dbError;
+  }
+};
+
+/**
+ * Handles vacation request rejection/cancellation.
+ */
```
Jdallos, line 66: Param docs and improve descripition

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
+      userId,
+      existingStartDate,
+      existingEndDate,
+      contractedWeek
+    );
+    await sendNotifications();
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
```
Jdallos, line 89: Same as previous comment, can we improve typing so this as unknown casting is not needed

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
+    };
+  } catch (keycloakError) {
+    console.error("Failed to return vacation days after status change.", keycloakError);
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
```
Jdallos, line 96: As above

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -3,17 +3,94 @@ import { middyfy } from "@libs/lambda";
...
+      existingEndDate,
+      contractedWeek
+    );
+    await sendNotifications();
+
+    return {
+      statusCode: 200,
+      body: JSON.stringify(updatedVacationRequest as unknown as VacationRequest)
+    };
+  } catch (keycloakError) {
```
Jdallos:
```suggestion
  } catch (error) {
```

src/libs/vacation-utils.ts
```diff
@@ -126,10 +141,41 @@ export const updateRemainingVacationDays = async (
   }
 
   user.unspentVacationDaysByYear = [...new Set(updatedUnspent)];
-  if (status === "APPROVED") {
-    await keycloakApiService.updateUserAttributes(userId, user);
+  await keycloakApiService.updateUserAttributes(userId, user);
+};
+
+/**
+ * Returns vacation days to Keycloak (adds them back).
```
Jdallos:
```suggestion
 * Returns vacation days to Keycloak user (adds them back).
```

src/libs/vacation-utils.ts
```diff
@@ -88,32 +88,47 @@ export const splitVacationDaysByYear = (
...
 
   const unspentVacationDaysByYear = user.unspentVacationDaysByYear || [];
+  const currentEntry = unspentVacationDaysByYear.find((s) => s.startsWith(`${year}:`));
+  const currentValue = currentEntry ? Number(currentEntry.split(":")[1]) : 0;
+
+  return currentValue >= daysNeeded;
+};
 
+/**
+ * Low-level helper that directly modifies Keycloak vacation day balances.
```
Jdallos:
```suggestion
 * Deducts vacation days from keycloak user
```

src/libs/vacation-utils.ts
```diff
@@ -126,10 +141,41 @@ export const updateRemainingVacationDays = async (
...
+
+  const unspentVacationDaysByYear = user.unspentVacationDaysByYear || [];
+  const currentEntry = unspentVacationDaysByYear.find((s) => s.startsWith(`${year}:`));
+  const currentValue = currentEntry ? Number(currentEntry.split(":")[1]) : 0;
+
+  const newRemainingDays = currentValue + daysToReturn;
+  const formattedValue = `${year}:${String(newRemainingDays).padStart(3, "0")}`;
+
+  const updatedUnspent = [...unspentVacationDaysByYear];
+  const existingIndex = updatedUnspent.findIndex((s) => s.startsWith(`${year}:`));
```
Jdallos: Improving the name here makes the conditions below clearer- its the existing index of what.
e.g.
yearsUnspentVacationDays

Maybe

Jdallos (approved):

src/functions/vacation-request/create-vacation-request/handler.ts
Dshad-Anon: made the suggested error code 409.

src/functions/vacation-request/update-vacation-request/handler.ts
Dshad-Anon: This function has database layer dependency and some HTTP responses,so I keep them in here and it also has sendNotificatoin which is tied to this handler sending notification. VacationUtil consist of vacationDays calculation so making it clear over there.

src/functions/vacation-request/update-vacation-request/handler.ts
Dshad-Anon: The type mismach is that the database model stores dates as string for dynamoDB while OpenAPI model expects Date objects. So double casting throuh unknown was used first to erase to unknown then applying target type.

src/libs/vacation-utils.ts
Dshad-Anon: made the changes.

src/functions/vacation-request/update-vacation-request/handler.ts
Jdallos: Then this works as a temporary fix but further work could be to resolve this type difference. I'd suggest looking into creating a VacationDTO type with some dtoToEntity and entityToDto methods to create a mapping layer allow handling of the type difference.
This is the correct way to handle type differences between API spec and DB.
I'll create a trello card, self assign if you want this task.

src/functions/vacation-request/update-vacation-request/handler.ts
Jdallos: Fine for now, this handler could be starting to get bloated and may benefit from a refactor, moving these and the notification functions into th eservice perhaps?
Again a potential future task in the trello

PR #323, opened 2026-01-16
Update Slack avatar Lambda to omit image_original when missing and include reason

JuusoHeikkinen02:
Made changes to the get-slack-avatar handler so that the response will return image_original if there is a value, if not return a reason why.

Added enums that describe the reason the image_original is not present

Also generated the spec to have access to the type-safe OpenAPI class for the ReasonEnums so there is package updates in the process.

### DIFF
--- a/src/functions/slack/get-slack-user-avatar/handler.ts
+++ b/src/functions/slack/get-slack-user-avatar/handler.ts
@@ -1,11 +1,12 @@
 import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
+import { SlackAvatarResponse } from "src/generated/homeLambdasModels/model/slackAvatarResponse";
 import SlackUtilities from "src/meta-assistant/slack/slack-utils";
 
 /**
  * Lambda to receive user's slack avatar by email
  *
  * @param event event with query parameter email
- * @return A response object with statuscode and avatar URL
+ * @return SlackAvatarResponse object with image URL or reason for failure
  */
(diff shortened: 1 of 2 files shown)


src/functions/slack/get-slack-user-avatar/handler.ts
```diff
@@ -21,20 +22,19 @@ const getSlackUserAvatarHandler: APIGatewayProxyHandler = async (event: APIGatew
...
 
-    const response = {
-      image_original: slackUser.profile.image_original
-    };
+    if (slackUser?.profile?.image_original) {
+      responseBody.imageOriginal = slackUser.profile.image_original;
+    } else {
+      responseBody.reason = slackUser
+        ? SlackAvatarResponse.ReasonEnum.NoAvatar
+        : SlackAvatarResponse.ReasonEnum.EmailMismatch;
```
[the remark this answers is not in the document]
JuusoHeikkinen02: Will implement if someone agrees with the change

Dshad-Anon (approved):

MariaKatariinaH (approved):

src/functions/slack/get-slack-user-avatar/handler.ts
Jdallos, line 32: I notice the spec had 3 strings within the reasonEnum, but is one of them unused?

Jdallos (approved):

src/functions/slack/get-slack-user-avatar/handler.ts
JuusoHeikkinen02: Was for like a default error, but I'll remove that one from the spec

PR #327, opened 2026-01-19
Bugfix-severa-env-email-lookup-in-opts

jessesillman:
## What has been done
- Ensured that the env is present for the severa user find.

### DIFF
--- a/src/functions/severa/add-opt-in/handler.ts
+++ b/src/functions/severa/add-opt-in/handler.ts
@@ -14,7 +14,7 @@ export const addSeveraOptInHandler: APIGatewayProxyHandler = async (event) => {
   if (!keycloakUserId) {
     return {
       statusCode: 400,
-      body: JSON.stringify({ error: "Keycloak userId is required" })
+      body: JSON.stringify({ message: "Keycloak userId is required" })
     };
   }
 
@@ -22,8 +22,9 @@ export const addSeveraOptInHandler: APIGatewayProxyHandler = async (event) => {
     const severaApi = CreateSeveraApiService();
     const keycloakApi = CreateKeycloakApiService();
 
-    // Fetch Keycloak user and email
-    let userEmail: string | undefined;
+    // Resolve Severa user by email
+    let severaUserId: string;
+
     try {
       const keycloakUser = await keycloakApi.findUser(keycloakUserId);
       if (!keycloakUser) {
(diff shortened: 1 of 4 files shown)


src/utils/severa/index.ts
```diff
@@ -10,8 +10,12 @@ import {CreateSeveraApiService} from "src/services/severa-api-service";
  *
  * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
  */
+export const getLookupEmail = (keycloakEmail: string): string | undefined => {
+  return process.env.SEVERA_TEST_USER_EMAIL ?? Config.get().testUser?.email ?? keycloakEmail;
```
JuusoHeikkinen02: Since the config uses the same env, would it make sense to rely on Config only and simplify this to:
```suggestion
  return Config.get().testUser?.email ?? keycloakEmail;
```

MariaKatariinaH (approved):

JuusoHeikkinen02 (approved):

src/functions/severa/add-opt-in/handler.ts
```diff
@@ -2,18 +2,19 @@ import type { APIGatewayProxyHandler } from "aws-lambda";
...
  */
 export const addSeveraOptInHandler: APIGatewayProxyHandler = async (event) => {
   const keycloakUserId = event.pathParameters?.userId;
   if (!keycloakUserId) {
-    return { 
-      statusCode: 400, 
-      body: JSON.stringify({ error: "Keycloak userId is required" }) 
+    return {
+      statusCode: 400,
+      body: JSON.stringify({ error: "Keycloak userId is required" })
```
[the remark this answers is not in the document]
Jdallos: It does seem we use message else where in the file so perhaps its right to use it here as well?

src/utils/severa/index.ts
```diff
@@ -10,8 +10,12 @@ import {CreateSeveraApiService} from "src/services/severa-api-service";
  *
  * @returns The user's Severa GUID, 'isSeveraOptIn' status, and email.
  */
+export const getLookupEmail = (keycloakEmail: string): string | undefined => {
```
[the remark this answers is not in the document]
Jdallos: This comment also seems correct and should be implemented

Jdallos, line 12: The TSDOcs for this getLookupEmail are outdated and should be updated

src/functions/severa/add-opt-in/handler.ts
```diff
@@ -26,25 +27,27 @@ export const addSeveraOptInHandler: APIGatewayProxyHandler = async (event) => {
...
-          body: JSON.stringify({ message: "Keycloak user not found." }) 
+        return {
+          statusCode: 404,
+          body: JSON.stringify({ message: "Keycloak user not found." })
         };
       }
 
-      // Use email from Keycloak user for Severa lookup
-      userEmail = keycloakUser.email;
+      // Use util to decide lookup email
```
Jdallos:
```suggestion
```

src/functions/severa/add-opt-in/handler.ts
Jdallos: Do we need to initialize the userEmail value here, why can't we just do it below as the response for getLoopupEmail e.g. 
 const userEmail = getLookupEmail(keycloakUser.email);

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -24,25 +25,36 @@ export const removeSeveraOptInHandler: APIGatewayProxyHandler = async (event) =>
 
     // Fetch Keycloak user and severaUserId attribute
     let severaUserId: string | undefined;
-    try {
+    let keycloakUserEmail: string | undefined;
```
Jdallos, line 28: Maybe I'm missing something, but does keycloakUserEmail actually need to be intialized here can't it be done inline below:
const keycloakUserEmail = keycloakUser.email;

Same for severaUserId as well?

src/functions/severa/remove-opt-in/handler.ts
```diff
@@ -24,25 +25,36 @@ export const removeSeveraOptInHandler: APIGatewayProxyHandler = async (event) =>
...
 
-    } catch {
-      return {
-        statusCode: 502,
-        body: JSON.stringify({ message: "Failed to fetch Keycloak user." }),
-      };
-    }
+      // If severaUserId missing, try resolving by email using shared util
+      if (!severaUserId) {
+        const lookupEmail = getLookupEmail(keycloakUserEmail);
```
Jdallos, line 43: Could name it just userEmail?

Jdallos (approved):

src/functions/severa/remove-opt-in/handler.ts
jessesillman: I changed the code so that the keycloakUserEmail is now inline inside the same try block, but i left severaUserId outside so it remanins available later in the code.

PR #338, opened 2026-03-02
Added permission limits for severa API calls.

### LINKED ISSUE #337: Refactoring Severa permission to Centralised master list.

Dshad-Anon:


MikaelPL1 (approved):

rashinikarunarathne (approved):

src/services/severa-api-service.ts
```diff
@@ -14,6 +14,55 @@ import type {
...
+ */
+export type SeveraPermission =
+  | "projects:read"
+  | "resourceallocations:read"
+  | "hours:read"
+  | "users:read"
+  | "users:write"
+  | "users:delete"
+  | "settings:write"
+  | "settings:read";
```
Jdallos: I wonder if it might be worth defining these strings as an enum, and then that can be referenced in both severaPermission and ALL_SEVERA_PERMISSIONS, it would prevent some future permission being added to one and not the other.

src/services/severa-api-service.ts
```diff
@@ -14,6 +14,55 @@ import type {
...
+
+/**
+ * Gets valid Severa permissions based on deployment stage
+ * Returns empty array in production, full permissions otherwise
+ *
+ * @returns Array of valid Severa permissions for current environment
+ */
+const getValidSeveraPermissions = (): ReadonlyArray<SeveraPermission> => {
+  const stage = process.env.STAGE?.toLowerCase();
+  const isProduction = stage === "production" || stage === "prod";
```
Jdallos: As far as I can tell this is a new env, at least it doesn't seem to exist in any of the hcv vaults. Do you have this locally or do you already get a value based on this env? Please confirm if this is the case and I can add it if needed.
I'd suggest the potential values should be "production" "staging" or "local" and it should be typed as such in the lambda env config?

Jdallos: And so you could then update this line as appropriate e.g. no || stage === "prod"

src/services/severa-api-service.ts
```diff
@@ -738,20 +787,60 @@ export const CreateSeveraApiService = (): SeveraApiService => {
...
 /**
- * Gets Severa access token
+ * Validates that all requested permissions are in the master list
+ *
+ * @param permissions - Array of permissions to validate
+ * @throws Error if any permission is not in the VALID_SEVERA_PERMISSIONS list
+ */
+const validateSeveraPermissions = (permissions: SeveraPermission[]): void => {
+  const stage = process.env.STAGE?.toLowerCase();
+  const isProduction = stage === "production" || stage === "prod";
```
Jdallos: See previous comment

serverless.ts
```diff
@@ -91,6 +91,7 @@ const serverlessConfiguration: AWS = {
     environment: {
       AWS_NODEJS_CONNECTION_REUSE_ENABLED: "1",
       NODE_OPTIONS: "--enable-source-maps --stack-trace-limit=1000",
+      STAGE: "${opt:stage, 'local'}",
```
Jdallos, line 94: Aha, well this answers the previous comment, but wouldn't it be better to handle this via HCV. This way might also be fine- your call.

Jdallos (approved):

src/services/severa-api-service.ts
Dshad-Anon: yes copilot did suggest the drift might occur when some additional permissions are added in one and left in another. Now I made it better than enum a const with single source of truth so it would prevent that future permissions.

src/services/severa-api-service.ts
Dshad-Anon: made the changes for it to be production only.

serverless.ts
Dshad-Anon: This is okay as its not a secret and also deploymentBucket is also made like that. So for easy deployment we can just change the --stage flag value. Also I made it so that it was easy to test production mode locally with --stage production.

src/services/severa-api-service.ts
Dshad-Anon: removed prod check

PR #350, opened 2026-03-11
Dto's for software registry

### LINKED ISSUE #349: DTO's for software registry

MikaelPL1:


src/database/models/software.ts
```diff
@@ -17,9 +17,9 @@ export interface SoftwareModel {
   image: string;
   status?: Status;
   createdBy: string;
-  createdAt?: Date;
+  createdAt: string;
```
Dshad-Anon: Why are the types changed from Date to string and also made from optional to necessary ones here?

src/functions/software-registry/create-software/handler.ts
```diff
@@ -47,25 +48,30 @@ export const createSoftwareHandler: ValidatedEventAPIGatewayProxyEvent<SoftwareR
     }
 
     const loggedUserId = authData.sub;
+    const now = new Date();
 
-    const newSoftware = {
+    const softwareDto = {
```
Dshad-Anon: newSoftware name is better here as we're creating a new software.

src/database/models/software.ts
MikaelPL1: Type 'string' is not assignable to type 'Date'. 
So if its date I can't use .toISOString

MikaelPL1: Changed back to Date and to optional, conversion still works.

arodriguezvaz (approved):

rashinikarunarathne (approved):

PR #369, opened 2026-03-23
lambda

### LINKED ISSUE #368: Lambda for import-document

MikaelPL1:
Lambda for import-document

### DIFF
--- a/src/functions/wiki-documentation/import-document/handler.ts
+++ b/src/functions/wiki-documentation/import-document/handler.ts
@@ -17,6 +17,7 @@ const responseHeaders = {
   "Access-Control-Allow-Origin": "*",
   "Access-Control-Allow-Credentials": true
 };
+
 /**
  * Converts a string into a URL-safe slug segment.
  * @param value Title value
--- a/src/functions/wiki-documentation/import-document/handler.ts
+++ b/src/functions/wiki-documentation/import-document/handler.ts
@@ -36,6 +36,7 @@ const importDocumentHandler: APIGatewayProxyHandler = async (event: APIGatewayPr
       statusCode: 500,
       body: JSON.stringify({
         code: 500,
+        headers: responseHeaders,
         message: "Invalid lambda environment variables"
       })
     };
--- a/src/functions/wiki-documentation/import-document/handler.ts
+++ b/src/functions/wiki-documentation/import-document/handler.ts
@@ -121,13 +121,6 @@ const importDocumentHandler: APIGatewayProxyHandler = async (event) => {
     }
 
     const bytes = await file.Body?.transformToByteArray();
-    if (!bytes?.length) {
-      return {
-        statusCode: 422,
-        headers: responseHeaders,
-        body: JSON.stringify({ message: "Empty PDF" })
-      };
-    }
 
     const markdown = await extractMarkdownFromPdf(new Uint8Array(bytes));
     if (!markdown) {
--- a/src/functions/wiki-documentation/import-document/handler.ts
+++ b/src/functions/wiki-documentation/import-document/handler.ts
@@ -138,11 +138,13 @@ const importDocumentHandler: APIGatewayProxyHandler = async (event: APIGatewayPr
       body: JSON.stringify({ path: basePath })
     };
   } catch (error) {
+    console.error("Error importing document", error);
     return {
       statusCode: 500,
       headers: responseHeaders,
       body: JSON.stringify({
-        message: error instanceof Error ? error.message : "Import failed"
+        code: "IMPORT_DOCUMENT_ERROR",
+        message: "Import failed"
       })
     };
   }
--- a/src/functions/wiki-documentation/import-document/handler.ts
+++ b/src/functions/wiki-documentation/import-document/handler.ts
@@ -1,77 +1,25 @@
 import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";
 import { middyfy } from "@libs/lambda";
-import type { APIGatewayProxyHandler } from "aws-lambda";
-import { PDFParse } from "pdf-parse";
+import type { APIGatewayProxyEvent, APIGatewayProxyHandler } from "aws-lambda";
 import { articlesApiService } from "src/database/services";
+import { responseHeaders } from "src/functions/wiki-documentation/upload-file/handler";
 import { getAuthDataFromToken } from "src/libs/auth-utils";
+import { extractMarkdownFromPdf, slugify, validateFileType } from "src/utils/importDocument";
 import { v4 as uuidv4 } from "uuid";
 
+/**
+ * Request payload for importing a document.
+ */
 type ImportDocumentRequest = {
-  Path?: string;
   path?: string;
   documentTitle?: string;
   overwriteExisting?: boolean;
 };
 
-const responseHeaders = {
-  "Access-Control-Allow-Origin": "*",
-  "Access-Control-Allow-Credentials": true
-};
-
 /**
- * Converts a string into a URL-safe slug segment.
- * 
- *
- * @param value Title value
+ * Lambda function for importing a PDF from s3 and creating a wiki article from it.
  */
-const slugify = (value: string) =>
-  value
-    .toLowerCase()
-    .trim()
-    .replaceAll(/[^a-z0-9\s-]/g, "")
-    .replaceAll(/\s+/g, "-");
-
-/**
- * Makes sure that uploaded file is PDF
- * 
- *
- * @returns API error response when invalid; otherwise `undefined`.
- */
-const validateFileType = (contentType: string | undefined, key: string) => {
-  const isPdf = contentType === "application/pdf" || key.toLowerCase().endsWith(".pdf");
-  if (!isPdf) {
-    return {
-      statusCode: 422,
-      headers: responseHeaders,
-      body: JSON.stringify({ message: "Only PDF files are allowed" })
-    };
-  }
-};
-
-/**
- * Extracts text from PDF bytes and normalizes it into markdown.
- *
- * @param bytes - PDF file content as bytes.
- * @returns Normalized markdown-like text content.
- */
-const extractMarkdownFromPdf = async (bytes: Uint8Array) => {
-  const parser = new PDFParse({ data: bytes });
-
-  try {
-    const parsed = await parser.getText();
-    return parsed.text
-      .split("\n")
-      .map((l) => l.trim())
-      .filter(Boolean)
-      .join("\n\n");
-  } finally {
-    await parser.destroy();
-  }
-};
-
-const importDocumentHandler: APIGatewayProxyHandler = async (event) => {
-  // ADMIN ONLY NEEDS TO BE ADDED LATER
-
+const importDocumentHandler: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {
   if (!event.body) {
     return {
       statusCode: 400,
@@ -109,8 +57,7 @@ const importDocumentHandler: APIGatewayProxyHandler = async (event) => {
     payload = event.body as unknown as ImportDocumentRequest;
   }
 
-  const path = payload.path || payload.Path;
-  const { documentTitle } = payload;
+  const { path, documentTitle } = payload;
 
   if (!path || !documentTitle) {
     return {
@@ -130,9 +77,9 @@ const importDocumentHandler: APIGatewayProxyHandler = async (event) => {
       })
     );
 
-    const validationResponse = validateFileType(file.ContentType, path);
-    if (validationResponse) {
-      return validationResponse;
+    const invalidFileTypeResponse = validateFileType(file.ContentType, path);
+    if (invalidFileTypeResponse) {
+      return invalidFileTypeResponse;
     }
 
     const bytes = await file.Body?.transformToByteArray();
@@ -177,7 +124,7 @@ const importDocumentHandler: APIGatewayProxyHandler = async (event) => {
       lastUpdatedBy: userId,
       lastUpdatedAt: now,
       lastReadAt: now,
-      readBy: [userId],
+      readBy: [],
       tags: [],
       draft: false,
       coverImage: undefined
--- a/src/functions/wiki-documentation/import-document/handler.ts
+++ b/src/functions/wiki-documentation/import-document/handler.ts
@@ -20,6 +20,8 @@ const responseHeaders = {
 
 /**
  * Converts a string into a URL-safe slug segment.
+ * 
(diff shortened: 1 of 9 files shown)


src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,205 @@
+import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";
+import { middyfy } from "@libs/lambda";
+import type { APIGatewayProxyHandler } from "aws-lambda";
+import { PDFParse } from "pdf-parse";
+import { articlesApiService } from "src/database/services";
+import { getAuthDataFromToken, isAdminUser } from "src/libs/auth-utils";
```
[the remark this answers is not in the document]
MikaelPL1: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Metatavu/home-lambdas/pull/369#discussion_r2974237897)

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,188 @@
...
+      lastUpdatedAt: now,
+      lastReadAt: now,
+      readBy: [userId],
+      tags: [],
+      draft: false,
+      coverImage: undefined
+    };
+
+    await articlesApiService.createArticle(article);
+
```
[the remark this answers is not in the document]
MikaelPL1: can be added later

rashinikarunarathne (approved):

arodriguezvaz (approved):

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
+import { GetObjectCommand, S3Client } from "@aws-sdk/client-s3";
+import { middyfy } from "@libs/lambda";
+import type { APIGatewayProxyHandler } from "aws-lambda";
+import { PDFParse } from "pdf-parse";
+import { articlesApiService } from "src/database/services";
+import { getAuthDataFromToken } from "src/libs/auth-utils";
+import { v4 as uuidv4 } from "uuid";
+
+type ImportDocumentRequest = {
```
Jdallos, line 12: All types should have some comments/ Docs even if it just states the name of the type. See other handlers for example. We want to be consistent throughtout the project

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+  Path?: string;
+  path?: string;
+  documentTitle?: string;
+  overwriteExisting?: boolean;
+};
+
+const responseHeaders = {
+  "Access-Control-Allow-Origin": "*",
+  "Access-Control-Allow-Credentials": true
+};
```
Jdallos:
```suggestion
};

```

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+import { v4 as uuidv4 } from "uuid";
+
+type ImportDocumentRequest = {
+  Path?: string;
+  path?: string;
+  documentTitle?: string;
+  overwriteExisting?: boolean;
+};
+
+const responseHeaders = {
```
Jdallos: I'm wondering if the exact same headers are used elsewhere e.g. the articles upload function. If they are this should be an exported from the types (or consts). If they are different leave as is

src/functions/wiki-documentation/import-document/handler.ts
Jdallos: line break

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+  "Access-Control-Allow-Origin": "*",
+  "Access-Control-Allow-Credentials": true
+};
+/**
+ * Converts a string into a URL-safe slug segment.
```
Jdallos:
```suggestion
 * Converts a string into a URL-safe slug segment.
 * 
 *
```

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+    .toLowerCase()
+    .trim()
+    .replaceAll(/[^a-z0-9\s-]/g, "")
+    .replaceAll(/\s+/g, "-");
+
+/**
+ * Makes sure that uploaded file is PDF
+ * @returns API error response when invalid; otherwise `undefined`.
+ */
+
```
Jdallos:
```suggestion
```

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+ */
+const slugify = (value: string) =>
+  value
+    .toLowerCase()
+    .trim()
+    .replaceAll(/[^a-z0-9\s-]/g, "")
+    .replaceAll(/\s+/g, "-");
+
+/**
+ * Makes sure that uploaded file is PDF
```
Jdallos:
```suggestion
 * Makes sure that uploaded file is PDF
 * 
 *
```

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+};
+
+/**
+ * Extracts text from PDF bytes and normalizes it into markdown.
+ *
+ * @param bytes - PDF file content as bytes.
+ * @returns Normalized markdown-like text content.
+ */
+
```
Jdallos:
```suggestion
```

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+      .split("\n")
+      .map((l) => l.trim())
+      .filter(Boolean)
+      .join("\n\n");
+  } finally {
+    await parser.destroy();
+  }
+};
+
+const importDocumentHandler: APIGatewayProxyHandler = async (event) => {
```
Jdallos: Missing TSDOcs

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+  // ADMIN ONLY NEEDS TO BE ADDED LATER
```
Jdallos:
```suggestion
  // TODO: ADMIN ONLY NEEDS TO BE ADDED LATER
```

You can install the TODO extension in VScode which adds some highlighting to TODO: comments

Jdallos: Although if this is regarding restriciting the feature so only admins can updload, doesn't this need to be done as part of this PR?

src/functions/wiki-documentation/import-document/handler.ts
Jdallos: I think most of these functions above the handler should be extracted into some shared util folder e.g. utils/importDocument or utils/pdf or similar

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  if (!event.body) {
```
Jdallos, line 22: Can we get stricter typing on this? We should have types available from the spec, which should describe the body the function will recieve from the UI.
You might need to update the generated type files from teh home lambdas spec to get access to these.
cd into the spec, git brach to check you are on develop and then git pull.
Then you can npm i which triggers the postinstall script OR you can directly run :
npm run build-home-lambdas-models

See the scripts in the package.json for more info

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+        headers: responseHeaders,
+        body: JSON.stringify({ message: "Invalid JSON" })
+      };
+    }
+  } else {
+    // When using middy/http-json-body-parser, event.body is already an object
+    payload = event.body as unknown as ImportDocumentRequest;
+  }
+
+  const path = payload.path || payload.Path;
```
Jdallos: Do we really need to handle both?

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+ * @param bytes - PDF file content as bytes.
+ * @returns Normalized markdown-like text content.
+ */
+
+const extractMarkdownFromPdf = async (bytes: Uint8Array) => {
+  const parser = new PDFParse({ data: bytes });
+
+  try {
+    const parsed = await parser.getText();
+    return parsed.text
```
Jdallos: Even if there were some issues running this locally were you able to try invoking the function (see read me) and logging the output of this?

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  try {
+    const file = await s3.send(
+      new GetObjectCommand({
+        Bucket: HOME_BUCKET_NAME,
+        Key: path
+      })
+    );
+
+    const validationResponse = validateFileType(file.ContentType, path);
```
Jdallos: Maybe rename this to make it explicit that validationResponse is really invalidFileType or similar?

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+      path: basePath,
+      title: documentTitle,
+      description: "Imported from PDF",
+      content: markdown,
+      createdBy: userId,
+      createdAt: now,
+      lastUpdatedBy: userId,
+      lastUpdatedAt: now,
+      lastReadAt: now,
+      readBy: [userId],
```
Jdallos: Not sure we should assume the uploading user has read the article, I see why you would add this but I think for our use case we should leave this empty.

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  const s3 = new S3Client({ region: HOME_BUCKET_REGION });
+
+  try {
+    const file = await s3.send(
+      new GetObjectCommand({
+        Bucket: HOME_BUCKET_NAME,
+        Key: path
+      })
+    );
```
Jdallos, line 74: So the event.body the handler is sent is not actually the PDF but  json with a url to a pdf that is already in the s3?
Well this does match the spec 👍 , but wasn't the plan to allow the user to upload a file/ pdf through the UI. So how is that being handled, otherwise how are we getting the PDF in the s3 in the first place?

Maybe I'm missing something here...

Jdallos (changes requested):

src/functions/wiki-documentation/import-document/handler.ts
MikaelPL1: Yes I tested it locally with a pdf I created.
returns the text correctly.

MikaelPL1: tested it in separate project

rashinikarunarathne (approved):

src/utils/importDocument/index.ts
```diff
@@ -0,0 +1,51 @@
+import { PDFParse } from "pdf-parse";
```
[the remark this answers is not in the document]
MikaelPL1: pdf-parse didn't work in this node version or / similiar problems was there when using it.

arodriguezvaz (approved):

arodriguezvaz (approved):

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -88,6 +36,7 @@ const importDocumentHandler: APIGatewayProxyHandler = async (event) => {
       statusCode: 500,
       body: JSON.stringify({
         code: 500,
+        headers: responseHeaders,
```
Jdallos: Do we really need to return the headers here? If we do please explain why

src/utils/importDocument/index.ts
```diff
@@ -0,0 +1,51 @@
...
+export const slugify = (value: string) =>
+  value
+    .toLowerCase()
+    .trim()
+    .replaceAll(/[^a-z0-9\s-]/g, "")
+    .replaceAll(/\s+/g, "-");
+
+/**
+ * Makes sure that uploaded file is PDF
+
```
Jdallos: Missing *

src/utils/importDocument/index.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export const validateFileType = (contentType: string | undefined, key: string) => {
+  const isPdf = contentType === "application/pdf" || key.toLowerCase().endsWith(".pdf");
+  if (!isPdf) {
+    return {
+      statusCode: 422,
+      headers: responseHeaders,
+      body: JSON.stringify({ message: "Only PDF files are allowed" })
+    };
+  }
```
Jdallos: Maybe it would be an improvement and more typical/ clearer to return a boolean from here and then directly return the object from the handler, rather than from here.

src/functions/wiki-documentation/import-document/handler.ts
```diff
@@ -1,77 +1,25 @@
...
 import { v4 as uuidv4 } from "uuid";
 
+/**
+ * Request payload for importing a document.
+ */
 type ImportDocumentRequest = {
-  Path?: string;
   path?: string;
   documentTitle?: string;
   overwriteExisting?: boolean;
```
Jdallos, line 15: Is this unused? Maybe this needs linking with the createArticle call

src/utils/importDocument/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  const parser = new PDFParse({ data: bytes });
+
+  try {
+    const parsed = await parser.getText();
+    return parsed.text
+      .split("\n")
+      .map((l) => l.trim())
+      .filter(Boolean)
+      .join("\n\n");
+  } finally {
```
Jdallos, line 41: Would it be worth having some catch log here incase it fails?

src/functions/wiki-documentation/import-document/handler.ts
Jdallos: I looked into this a little and I'm now skeptical.
We seem to be using them inconsistently - not for all responses.
This is used throughout upload but then in other functions we don't use it at all.

I suspect we don't need to return these headers at all.
To begin with remove all from import document (unless you disagree) and then we can see if we get CORs issues once we deploy the lambda. IF we don't I'd say another small task would be to try removing teh headers from the upload-file function as well. But lets do that as a seperate PR if this works as intended.

Jdallos (approved):

PR #371, opened 2026-03-25
Feature/8 user status backend

rashinikarunarathne:
implemented an API endpoint to update the user isActive status. and set up to update keycloak attribute. when testing, got two issues JSON parsing and value format, but fixed. finally tested severless offline .


arodriguezvaz (approved):

MikaelPL1 (approved):

MikaelPL1:
Approved, check copilots comments.

src/functions/users/flextime/handler.ts
```diff
@@ -1,101 +1,121 @@
...
-    attributes: {
-      severaUserId: string;
-      isActive: boolean;
-    };
-  };
-  flextime: {
-    totalFlextimeBalance: number;
-    monthFlextimeBalance: number;
-  };
+	user: {
```
Jdallos: Biome/ linter issue, we should be working with spaces set to 2, this looks wrong and shows more changes in the PR making it harder to see what has actually changed.
Please ensure linter is set up and spaces set to 2 and you have format on save set up.

serverless.ts
```diff
@@ -57,6 +58,323 @@ const isLocal = process.env.STAGE === "local";
...
+						},
+					],
+					ProvisionedThroughput: {
+						ReadCapacityUnits: 1,
+						WriteCapacityUnits: 1,
+					},
+				},
+			},
+		},
+	},
```
Jdallos: I have no idea why all of this is showing up as change, maybe it's linter related?

src/functions/users/status/handler.ts
```diff
@@ -0,0 +1,50 @@
+import type { APIGatewayProxyHandler } from "aws-lambda";
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import { middyfy } from "src/libs/lambda";
+
+export const updateUserStatusHandler: APIGatewayProxyHandler = async (
```
Jdallos: Missing TSDocs

src/functions/users/status/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+		const isActive = body.isActive;
+
+		if (!userId || typeof isActive !== "boolean") {
+			return {
+				statusCode: 400,
+				body: JSON.stringify({ message: "Invalid request" }),
+			};
+		}
+
+		const keycloakApi = CreateKeycloakApiService();
```
Jdallos: We typically only need to wrap async functions in try catch blocks, so all that I have highlighted here doesn't need to be in the try, let me know if you disagree

src/functions/users/flextime/handler.ts
```diff
@@ -1,101 +1,121 @@
...
-    return {
-      statusCode: 500,
-      body: JSON.stringify({
-        code: 500,
-        message: "Internal server error",
-        error: error instanceof Error ? error.message : "Unknown error"
-      })
-    };
-  }
+	try {
```
Jdallos: Not reviewing this yet as can't tell what has changed and what is just linter related changes

Jdallos (changes requested):

src/functions/users/flextime/handler.ts
```diff
@@ -52,7 +60,7 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async () => {
               email: severaUser.email || "",
               attributes: {
                 severaUserId: severaUser.guid,
-                isActive: true
+                isActive: isActive || false
```
Jdallos: isActive is already a boolean so || false is unnecessary and can be removed

src/functions/users/status/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import { middyfy } from "src/libs/lambda";
+
+/**
+ *
+ * updates user's active status in keycloak.
+ * @param event - API gateway event contaning unseId and isActive flag
+ * @returns API response indicating sucess or failure
+ */
+
```
Jdallos:
```suggestion
```
Remove uinnecessary line break

src/functions/users/flextime/handler.ts
Jdallos, line 1: For the future tegarding branch name it has a typo, should be :
feature-8-user-status-backend
not
feature/8-user-status-backend

Also branch names should be descriptive, but also we know it's backend based on the repo so that's unnecessary.
As an example I would have named this:
feature-8-update-keycloak-user-isActive-status

Jdallos: Also the issue number is incorrect, I don't know where you got 8 from but this is not the 8th issue, current issue numbers are over 350 in this repo.
As a reminder you should create an issue in the issues tab for the repo and then use whatever number is assigned. If in doubt please check with another developer trainee

src/functions/users/status/handler.ts
```diff
@@ -0,0 +1,50 @@
...
+export const updateUserStatusHandler: APIGatewayProxyHandler = async (event) => {
+  const userId = event.pathParameters?.userId;
+  const body = event.body as unknown as { isActive: boolean };
```
Jdallos: We don't want to do this "as unknown" casting, when possible.
First the body seems to be being typed as a string intially as in other files (e.g. software update) you need to parse e.g. :
    const data: SoftwareRegistry =
      typeof event.body === "string" ? JSON.parse(event.body) : (event.body as SoftwareRegistry);

So for you maybe something like this would be appropriate:
  const body = event.body ? JSON.parse(event.body) : null;

Did you manaage to run this function via serverless offline or invoke the function (see readme). even if the functino were to crash at the keycloak request point (it might) you should still be able to log the body to determine what is needed here.

src/functions/users/update-status/handler.ts
Jdallos, line 1: The status folder should be renamed to update-status to be consistent with other functions in the project.

serverless.ts
```diff
@@ -33,7 +34,6 @@ import createArticleHandler from "src/functions/wiki-documentation/create-articl
 import deleteArticleHandler from "src/functions/wiki-documentation/delete-article";
 import findArticleHandler from "src/functions/wiki-documentation/find-article";
 import findArticleByPathHandler from "src/functions/wiki-documentation/find-article-by-path";
-import importDocument from "src/functions/wiki-documentation/import-document";
```
Jdallos, line 36: I think this is showing as removed as you are now behind develop.
When you don't have any local work to commit you should run 
git fetch
git merge origin/develop

To pull changes, and also make sure you are up to date with latest spec ref
So in the terminal cd into the spec from the home lambda root,
git branch to check you are on develop and then git pull

Jdallos (changes requested):

src/functions/users/flextime/handler.ts
```diff
@@ -44,6 +47,11 @@ export const listUsersFlextimeHandler: APIGatewayProxyHandler = async () => {
         try {
           // This should be typed according to the spec response type from the severa general spec generated client
```
Jdallos: This comment can now be removed as it looks like flextime is indeed correctly typed now

src/functions/users/flextime/handler.ts
```diff
@@ -25,14 +26,16 @@ type UserWithFlextime = {
...
  * Lambda handler for listing users with their flextime data.
  * Only returns users who have opted in to Severa integration.
- * 
+ *
  * @param event - API Gateway proxy event containing query parameters
  * @returns Promise resolving to API Gateway proxy result with user flextime data
  */
 export const listUsersFlextimeHandler: APIGatewayProxyHandler = async () => {
   try {
     const api = CreateSeveraApiService();
```
Jdallos: renaming this to severaApi could be good as we now have both the severa and keycloak api in use in this handler- so helps with clarity

src/functions/users/update-status/handler.ts
```diff
@@ -0,0 +1,49 @@
+import type { APIGatewayProxyHandler } from "aws-lambda";
+import { CreateKeycloakApiService } from "src/database/services/keycloak-api-service";
+import { middyfy } from "src/libs/lambda";
+
+/**
+ *
+ * updates user's active status in keycloak.
+ * @param event - API gateway event contaning unseId and isActive flag
+ * @returns API response indicating sucess or failure
+ */
```
Jdallos, line 11: /**
 * Updates user's active status in keycloak.
 * 
 * @param event - API gateway event contaning unseId and isActive flag
 * @returns API response indicating sucess or failure
 */

This is the format for TSDocs we want, with a * line break between the description and the @params

Jdallos: Please fix the multiple typos

Jdallos (approved):

PR #392, opened 2026-04-28
Update vacation status when someone is on vacation

arodriguezvaz:
Every day, checks if there are vacation requests that should be active or ended, and updates their status to ONGOING or COMPLETED accordingly.
Regarding updatedAt: new Date().toISOString() as unknown as Date — there is likely a better solution than this cast, but I couldn't find one


rashinikarunarathne (approved):

NeonValtokari (approved):

src/functions/vacation-request/update-vacation-status-by-date/handler.ts
```diff
@@ -0,0 +1,112 @@
...
+              createdBy: "system",
+              updatedAt
+            }
+          ],
+          updatedAt: updatedAt.toISOString()
+        });
+
+        return nextStatus;
+      })
+    );
```
Jdallos, line 90: It's a reasonable chunk of logic in this map.
I'd suggest either adding just a couple of comments to make clear what is happening in the various sections e.g. Checking dates to see if status should be updated.
OR you could refactor into a couple of smaller functions e.g. 
checkNextStatus might be a reasonable candidate for a function

OR leave as is is also ok.

src/functions/vacation-request/update-vacation-status-by-date/handler.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    for (const result of results) {
+      if (result.status === "rejected") {
+        errors.push(result.reason instanceof Error ? result.reason.message : String(result.reason));
+      } else if (result.value === VacationRequestStatuses.Ongoing) {
+        ongoingUpdated += 1;
+      } else if (result.value === VacationRequestStatuses.Completed) {
+        completedUpdated += 1;
+      }
```
Jdallos, line 102: Another candidate for a named function perhaps. Or as is is fine

src/functions/vacation-request/update-vacation-status-by-date/index.ts
```diff
@@ -0,0 +1,14 @@
+import { handlerPath } from "@libs/handler-resolver";
+
+const { DAILY_SCHEDULE_TIMER } = process.env;
```
Jdallos, line 3: Does this env already exists, and daily is the correct interval for this to run?

src/functions/vacation-request/update-vacation-status-by-date/handler.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
+const updateVacationStatusByDateHandler = async (): Promise<APIGatewayProxyResult> => {
+  try {
+    const today = DateTime.now().toISODate();
+
+    if (!today) {
+      throw new Error("Could not resolve current date");
+    }
+
+    const vacations = await vacationRequestService.listVacationRequests();
```
Jdallos: Is this listing all vacation requests, for all time? Is there some way this could be a little more targeted, otherwise after some years we might end up with quite a lot of results unnecessarily.

Jdallos (approved):

PR #394, opened 2026-04-29
Add vacation date validation for creating and updating requests

arodriguezvaz:
**Before merging is important to update the api-spec in this repository**

### DIFF
--- a/src/functions/vacation-request/update-vacation-request/handler.ts
+++ b/src/functions/vacation-request/update-vacation-request/handler.ts
@@ -8,7 +8,8 @@ import {
   getContractedWeek,
   getLatestStatus,
   returnVacationDaysForRejection,
-  validateVacationApproval
+  validateVacationApproval,
+  validateVacationDates
 } from "src/libs/vacation-utils";
 import {
   notifyAdminsVacationSubmittedAll,
@@ -162,6 +163,13 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
   const statusChanged = existingLatestStatus !== newLatestStatus;
   const draftStatusChanged = existingVacationRequest.draft !== draft;
 
+  const existingDays = existingVacationRequest.days;
+  const daysChanged = existingDays !== days;
+
+  const existingStartDate = existingVacationRequest.startDate;
+  const existingEndDate = existingVacationRequest.endDate;
+  const datesChanged = existingStartDate !== startDate || existingEndDate !== endDate;
+
   const vacationRequestUpdates = {
     id: existingVacationRequest.id,
     userId: existingVacationRequest.userId,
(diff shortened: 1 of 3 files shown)


NeonValtokari (approved):

rashinikarunarathne (approved):

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -201,30 +202,40 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
...
-        userId,
-        startDate,
-        endDate,
-        contractedWeek,
-        vacationRequestUpdates,
-        sendNotifications
-      );
-    }
+    const checkedDates = await validateVacationDates(startDate, endDate);
+    if (checkedDates) {
```
Jdallos: Might be the clearer patter to check for !checkedDates and have the appropriate return first. Then have the nested ifs below it, so they don't need to be nested.

Less nested condition is always a good thing. Also that's how its done in the other file updated in this PR- create vacation request handler

src/functions/vacation-request/update-vacation-request/handler.ts
```diff
@@ -201,30 +202,40 @@ const updateVacationRequestHandler: ValidatedEventAPIGatewayProxyEvent<
...
-        existingVacationRequest.startDate,
-        existingVacationRequest.endDate,
-        contractedWeek,
-        vacationRequestUpdates,
-        sendNotifications
-      );
+      if (
+        existingLatestStatus === VacationRequestStatuses.Approved &&
+        statusChanged &&
+        currentStatus !== VacationRequestStatuses.Approved
```
Jdallos: Could maybe create a named variable for these 3 conditions e.g. statusChangeToRejected.
Mainly for greater clarity at a clance.

 But it's also quite fine as is.

src/libs/vacation-utils.ts
```diff
@@ -109,7 +109,22 @@ export const validateVacationDays = async (
 
   return currentValue >= daysNeeded;
 };
+/**
+ * Checks if the start and end dates of a vacation request are valid (not in the past, start before end).
+ * @param startDate
+ * @param endDate
+ * @returns
```
Jdallos:
```suggestion
```

src/libs/vacation-utils.ts
```diff
@@ -109,7 +109,22 @@ export const validateVacationDays = async (
 
   return currentValue >= daysNeeded;
 };
+/**
+ * Checks if the start and end dates of a vacation request are valid (not in the past, start before end).
+ * @param startDate
+ * @param endDate
```
Jdallos:
```suggestion
 * @param startDate string
 * @param endDate string
```

src/libs/vacation-utils.ts
```diff
@@ -109,7 +109,22 @@ export const validateVacationDays = async (
 
   return currentValue >= daysNeeded;
 };
+/**
+ * Checks if the start and end dates of a vacation request are valid (not in the past, start before end).
```
Jdallos, line 114:
```suggestion
 * Checks if the start and end dates of a vacation request are valid (not in the past, start before end).
 *
```

src/libs/vacation-utils.ts
```diff
@@ -109,7 +109,22 @@ export const validateVacationDays = async (
...
+ * @param startDate
+ * @param endDate
+ * @returns
+ */
+export const validateVacationDates = async (
```
Jdallos: Does this function really need to be async, and does it actually return a promise?

Jdallos (approved):

